PR #10, opened 2026-08-18
feat: add iam clone policy tool

### LINKED ISSUE #3: Policies: no clone/duplicate tool (API supports it)
From [docs/capability-gap-analysis.md](../blob/main/docs/capability-gap-analysis.md), gap analysis 2026-08-18.

Scaleway's IAM API supports a policy CLONE operation (also exposed in the console as a "Duplicate" action), letting you copy an existing policy's rules into a new one in one call. This server has no `scaleway_iam_clone_policy` tool.

**Not blocking** - fully achievable today by chaining existing tools (`scaleway_iam_get_policy` + `scaleway_iam_list_policy_rules` + `scaleway_iam_create_policy`), just not atomic. Low priority - a convenience wrapper, not a missing capability.

gkrost:
Closes #3.

Adds \scaleway_iam_clone_policy\ wrapping the IAM CLONE operation (\POST /policies/{policy_id}/clone\, verified against scaleway-sdk-go). Atomic copy of rules/principal/tags; the clone's name is duplicated (no override fields on the API), so the tool description points to \scaleway_iam_update_policy\ for renaming.

Also marks the policy-clone gap as fixed in \docs/capability-gap-analysis.md\.

Build verified green with \
pm run build\.

### DIFF
--- a/src/tools/policies.ts
+++ b/src/tools/policies.ts
@@ -85,6 +85,10 @@ const updateSchema = {
     ),
 };
 
+const cloneSchema = {
+  policy_id: z.string().uuid(),
+};
+
 const listSchema = {
   application_id: z.string().uuid().optional().describe("Filter to policies attached to one Application."),
 };
@@ -141,6 +145,28 @@ export function registerPolicies(server: McpServer, config: Config) {
           group_id,
           rules,
         });
+        return toolJsonResult(policy, config.MAX_OUTPUT_CHARS);
+      }),
+  );
+
+  server.registerTool(
+    "scaleway_iam_clone_policy",
+    {
+      title: "Clone Scaleway IAM Policy",
+      description:
+        "Clone an existing IAM Policy in one atomic call (the API's CLONE operation, the console's " +
+        "'Duplicate' action). The new policy is an exact copy: same rules, same principal " +
+        "(application_id/user_id/group_id), same description and tags. Scaleway's clone API takes NO override " +
+        "fields - the clone's name is identical to the source, so rename it afterward with " +
+        "scaleway_iam_update_policy if a distinct name is needed (nothing references a Policy by name, so " +
+        "renaming is always safe). Prefer this over the manual get+create chain: it copies rules in one step " +
+        "without a window where the copy is only partially built.",
+      inputSchema: cloneSchema,
+      annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: false, openWorldHint: true },
+    },
+    async ({ policy_id }) =>
+      withIamErrorHandling(async () => {
+        const policy = await iamRequest<Policy>(config, "POST", `/policies/${policy_id}/clone`, {});
         return toolJsonResult(policy, config.MAX_OUTPUT_CHARS);
       }),
   );
--- a/src/tools/policies.ts
+++ b/src/tools/policies.ts
@@ -154,20 +154,31 @@ export function registerPolicies(server: McpServer, config: Config) {
     {
       title: "Clone Scaleway IAM Policy",
       description:
-        "Clone an existing IAM Policy in one atomic call (the API's CLONE operation, the console's " +
-        "'Duplicate' action). The new policy is an exact copy: same rules, same principal " +
-        "(application_id/user_id/group_id), same description and tags. Scaleway's clone API takes NO override " +
-        "fields - the clone's name is identical to the source, so rename it afterward with " +
-        "scaleway_iam_update_policy if a distinct name is needed (nothing references a Policy by name, so " +
-        "renaming is always safe). Prefer this over the manual get+create chain: it copies rules in one step " +
-        "without a window where the copy is only partially built.",
+        "Clone an existing IAM Policy. The result is an exact copy: same rules, same principal " +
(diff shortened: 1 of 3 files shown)


src/tools/policies.ts
```diff
@@ -145,6 +149,28 @@ export function registerPolicies(server: McpServer, config: Config) {
       }),
   );
 
+  server.registerTool(
+    "scaleway_iam_clone_policy",
+    {
+      title: "Clone Scaleway IAM Policy",
+      description:
+        "Clone an existing IAM Policy in one atomic call (the API's CLONE operation, the console's " +
+        "'Duplicate' action). The new policy is an exact copy: same rules, same principal " +
```
gkrost: **Confirmed false, live.** The clone never keeps the source's principal - it always comes back with `no_principal: true`. Tested twice, including explicitly passing `application_id` in the clone request body: silently ignored, still `no_principal: true`.

This isn't just a doc bug - `scaleway_iam_update_policy` (this file) only accepts `name`/`description`/`tags`, never `application_id`/`user_id`/`group_id`, so there's currently no way to reattach a principal to the clone through this server at all. I checked whether that's a Scaleway limitation: it isn't - `PATCH /policies/{id}` does accept and apply `application_id` live (tested directly against the raw API, 200 + principal changed). So the API supports the fix, this server's tool schema just doesn't expose it yet.

gkrost (commented):
Requesting changes (posted as COMMENT, not REQUEST_CHANGES - GitHub blocks that on your own PR). Live-tested this end to end against the real account before reviewing (create app -> create policy w/ a rule -> tag it -> clone -> compare -> cleanup, everything created was deleted afterward). Implementation itself is fine (correctly calls POST /policies/{id}/clone) - the problems are in what the description claims. See inline comments for the three specific wrong claims and the follow-up fix needed at the call site.

src/tools/policies.ts
```diff
@@ -145,6 +149,28 @@ export function registerPolicies(server: McpServer, config: Config) {
...
+        "(application_id/user_id/group_id), same description and tags. Scaleway's clone API takes NO override " +
```
gkrost: **"same description and tags" is half true.** `description` is copied correctly. `tags` are not - original had `["managed-by=pr10-review"]`, the clone came back `[]`.

src/tools/policies.ts
```diff
@@ -145,6 +149,28 @@ export function registerPolicies(server: McpServer, config: Config) {
...
+        "fields - the clone's name is identical to the source, so rename it afterward with " +
```
gkrost: **"the clone's name is identical to the source" is false.** Scaleway auto-prefixes it: cloning `pr10-review-1787049437380` produced `Copy of pr10-review-1787049437380`. Also confirmed an explicit `name` override in the request body is silently ignored, same as `application_id` above - the endpoint really does take zero effective overrides, it's just that the *default* isn't "identical name," it's "identical name with a prefix."

src/tools/policies.ts
```diff
@@ -145,6 +149,28 @@ export function registerPolicies(server: McpServer, config: Config) {
...
+        "scaleway_iam_update_policy if a distinct name is needed (nothing references a Policy by name, so " +
+        "renaming is always safe). Prefer this over the manual get+create chain: it copies rules in one step " +
+        "without a window where the copy is only partially built.",
+      inputSchema: cloneSchema,
+      annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: false, openWorldHint: true },
+    },
+    async ({ policy_id }) =>
+      withIamErrorHandling(async () => {
+        const policy = await iamRequest<Policy>(config, "POST", `/policies/${policy_id}/clone`, {});
```
gkrost: This is the actual fix needed (not just the description above): since the tool's premise is "exact copy," make it one. After this POST, `PATCH /policies/{policy_id}` (fetch the source first, e.g. via `iamRequest` GET on `/policies/{policy_id}`) with the source's `application_id`/`user_id`/`group_id` and `tags` before returning - that delivers what's promised instead of handing back an orphaned, tag-less policy. Rename only if the caller wants a specific name; Scaleway's own `Copy of X` default is fine otherwise.

While in here: also worth carrying forward the `nb_rules: 0`-in-the-immediate-response quirk this endpoint has (same eventual-consistency gotcha `docs/gotchas.md` already documents for `create_policy` - confirmed clone has it too), so the response here isn't misread as "rules didn't copy."

gkrost:
Pushed \`408e9a1\` addressing the review: went with option (b) - \`clone_policy\` now fetches the source, clones, then \`PATCH\`es the source's \`application_id\`/\`user_id\`/\`group_id\`/\`tags\` onto the result before returning, instead of handing back an orphaned policy.

Re-verified live end to end (source policy tagged + attached to an app, cloned, all fields checked, everything cleaned up):

\`\`\`
PASS: clone id != orig id
PASS: clone application_id == orig application_id
PASS: clone tags == orig tags
PASS: clone description == orig description
PASS: clone name starts with 'Copy of'
PASS: clone rules match original
\`\`\`

Also documented both gotchas this surfaced in \`docs/gotchas.md\`: the clone endpoint silently ignoring its request body and stripping principal/tags, and the shared \`nb_rules\`-undercounts-immediately quirk (confirmed it also affects \`create_policy\`, not just clone).

PR #11, opened 2026-08-18
feat(audit-trail): cover remaining Audit Trail API surface

### LINKED ISSUE #9: Audit Trail: alert rules, export jobs, and other endpoints beyond list-events not covered
From [docs/capability-gap-analysis.md](../blob/main/docs/capability-gap-analysis.md), gap analysis 2026-08-18.

`scaleway_audit_list_events` covers `GET .../events` only. Scaleway's Audit Trail API is considerably larger: authentication-events, system-events, combined-events, last-events-overview, export jobs (ship logs to storage), and - the one with real standalone value here - **alert rules + custom alert rules** (list/enable/disable/create/update/delete). Alerting is a genuinely different capability from querying (proactive vs. reactive), not just more of the same data.

Unlike the other gap issues filed today, this one isn't a clear scope boundary - if Audit Trail usage grows beyond ad-hoc "what happened" queries into "notify me when X happens," alert rules would be the natural next tool to add. Lower priority than the two gaps fixed today (`update_api_key`, bucket lifecycle) since nothing in this session's live re-test hit it directly - worth a look if/when it does.

gkrost:
Closes #9

## What

17 new Audit Trail tools covering the full documented API surface beyond \GET /events\:

**Event queries (read-only, \AuditTrailReadOnly\)**
- \scaleway_audit_list_authentication_events\ - login/token/MFA outcomes (separate endpoint)
- \scaleway_audit_list_system_events\ - Scaleway-initiated actions on your resources
- \scaleway_audit_list_combined_events\ - all three streams in one chronological feed, each event tagged api/auth/system
- \scaleway_audit_get_last_events_overview\ - the console landing-page snapshot
- \scaleway_audit_list_products\ - the valid filter-value catalog (products/services/methods)

**Preconfigured alert rules** (verified live)
- \scaleway_audit_list_alert_rules\, \scaleway_audit_set_alert_rules_enabled\ (additive), \scaleway_audit_replace_enabled_alert_rules\ (full-replace, confirm-guarded)

**Custom alert rules** (implemented; blocked on Scaleway - see findings)
- full CRUD + additive enable/disable + full-replace: 6 tools

**Export jobs** (verified live)
- \scaleway_audit_list_export_jobs\, \scaleway_audit_create_export_job\, \scaleway_audit_delete_export_job\ (confirm-guarded)

## How

- \uditClient.ts\: now supports POST/PATCH/PUT/DELETE with JSON bodies
- \uditTrailShared.ts\: shared schemas, error handling, cursor + numbered pagination, per-kind event compaction
- Smoke test extended with all new audit tools (tolerant read probes + custom-rule lifecycle)
- README + \docs/capability-gap-analysis.md\ updated

Live-verification findings posted as PR comment (key one: Scaleway's custom-alert-rules API is documented but unimplemented - HTTP 501).

### DIFF
--- a/src/tools/auditTrailShared.ts
+++ b/src/tools/auditTrailShared.ts
@@ -4,7 +4,7 @@ import { AuditTrailApiError } from "../auditClient.js";
 import { toolError } from "../output.js";
 
 export const auditRegionSchema = z
-  .enum(["fr-par", "nl-ams"])
+  .enum(["fr-par", "nl-ams", "pl-waw"])
   .optional()
   .describe("Defaults to the server's configured region (fr-par).");
 
--- a/src/tools/auditTrailShared.ts
+++ b/src/tools/auditTrailShared.ts
@@ -1,0 +1,206 @@
+import { z } from "zod";
+import type { Config } from "../config.js";
+import { AuditTrailApiError } from "../auditClient.js";
+import { toolError } from "../output.js";
+
+export const auditRegionSchema = z
+  .enum(["fr-par", "nl-ams"])
+  .optional()
+  .describe("Defaults to the server's configured region (fr-par).");
+
+export const recordedAfterSchema = z
+  .string()
+  .optional()
+  .describe("ISO 8601 date-time, inclusive. Defaults to 24 hours ago.");
+
+export const recordedBeforeSchema = z
+  .string()
+  .optional()
+  .describe("ISO 8601 date-time, exclusive. Defaults to now (omitted from the request).");
+
+export const maxPagesSchema = z
+  .number()
+  .int()
+  .min(1)
+  .max(50)
+  .optional()
+  .describe("Safety cap on pagination (100 items/page). Default 20. Said to be truncated in the response if more pages remained.");
+
+export const confirmSchema = z
+  .literal(true)
+  .describe("Must be explicitly true. Guards a destructive or implicitly-disabling operation.");
+
+export const WRITE_PERMISSIONS_NOTE =
+  "Needs more than AuditTrailReadOnly on THIS server's own credential (read-only covers just the " +
+  "query tools) - if the call fails with permissions_denied, grant the Audit Trail write permission " +
+  "set via scaleway_iam_set_policy_rules.";
+
+export async function handleAudit<T>(fn: () => Promise<T>): Promise<T | ReturnType<typeof toolError>> {
+  try {
+    return await fn();
+  } catch (err) {
+    if (err instanceof AuditTrailApiError) return toolError(err.message) as unknown as T;
+    throw err;
+  }
+}
+
+export function defaultRecordedAfter(): string {
+  return new Date(Date.now() - 24 * 60 * 60 * 1000).toISOString();
+}
+
+export interface AuditResource {
+  type: string;
+  id: string;
+  name?: string | null;
+}
+
+export function trimResource(r: AuditResource): AuditResource {
+  const out: AuditResource = { type: r.type, id: r.id };
+  if (r.name) out.name = r.name;
+  return out;
+}
+
+export interface ApiAuditEvent {
+  id: string;
+  recorded_at: string | null;
+  principal: { id: string };
+  organization_id: string;
+  project_id?: string | null;
+  source_ip: string;
+  user_agent?: string | null;
+  product_name: string;
+  service_name: string;
+  method_name: string;
+  resources: AuditResource[];
+  request_id: string;
+  request_body?: unknown;
+  status_code: number;
+}
+
+export function compactApiEvent(e: ApiAuditEvent): ApiAuditEvent {
+  return {
+    id: e.id,
+    recorded_at: e.recorded_at,
+    principal: e.principal,
+    product_name: e.product_name,
+    service_name: e.service_name,
+    method_name: e.method_name,
+    resources: (e.resources ?? []).map(trimResource),
+    status_code: e.status_code,
+  } as ApiAuditEvent;
+}
+
+export interface AuthAuditEvent {
+  id: string;
+  recorded_at: string | null;
+  organization_id: string;
+  source_ip: string;
+  user_agent?: string | null;
+  resources: AuditResource[];
+  result?: string | null;
+  failure_reason?: string | null;
+  country_code?: string | null;
+  method?: string | null;
+  origin?: string | null;
+  mfa_type?: string | null;
+}
+
+export function compactAuthEvent(e: AuthAuditEvent): AuthAuditEvent {
+  return {
+    id: e.id,
+    recorded_at: e.recorded_at,
+    source_ip: e.source_ip,
+    result: e.result,
+    failure_reason: e.failure_reason,
+    country_code: e.country_code,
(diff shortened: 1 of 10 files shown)


gkrost:
## Live-verification findings (2026-08-18, fr-par, org `ad82477a`)

All evidence below is from this session: the extended `scripts/smoke-test.mjs` against the built server, plus stateless HTTP probes (bogus-UUID / empty-body requests — 4xx proves the endpoint exists and validates; 501 proves it doesn't exist; unknown-method routing answers 501 before validation ever runs, which is what makes the distinction reliable).

### ✅ Verified working end-to-end (via MCP tools)

| Tool | Result |
|---|---|
| `scaleway_audit_list_authentication_events` | 200, `count: 0` (no auth events in window — empty, not broken) |
| `scaleway_audit_list_system_events` | 200, `count: 0` (same) |
| `scaleway_audit_list_combined_events` | 200, **59–67 events**, `kind` tags (`api`/`auth`/`system`) correct, compaction per kind works |
| `scaleway_audit_get_last_events_overview` | 200, 10 events, matches console landing page shape |
| `scaleway_audit_list_products` | 200, **22 products** with services/methods — this is the catalog for `resource_type`/`method_name` filters |
| `scaleway_audit_list_alert_rules` | 200, **11 preconfigured rules** (e.g. `ApiKeyCreated`), `total_count` correct; 0 currently enabled |
| `scaleway_audit_list_export_jobs` | 200, `total_count: 0` |

### ⚠️ Key finding: custom-alert-rules are documented but NOT implemented by Scaleway

Every custom-alert-rules method returns:

```
HTTP 501 {"message":"unknown method ListCustomAlertRules for service scaleway.audit_trail.v1alpha1.Api"}
```

Probed: `GET /custom-alert-rules`, `POST /custom-alert-rules`, `PUT /custom-alert-rules` (`SetEnabledCustomAlertRules`) — all 501. This is **not** a permissions problem (the preconfigured-rules endpoints on the same API work with the same credential). The tools are kept and will work when Scaleway ships the backend; each tool description carries this caveat verbatim. Tracked in `docs/capability-gap-analysis.md`.

### ✅ Alert-rule mutation endpoints verified statelessly (not exercised for real)

- `POST /enable-alert-rules`, `POST /disable-alert-rules`, `PATCH /alert-rules` (full-replace): bogus rule UUID → **404 `resource: alert_rules`, no partial application** — the whole request is rejected atomically. Real rule IDs were deliberately NOT toggled to avoid changing live alert state.
- `POST /export-jobs` with empty body → 400 with field-level validation details (`organization_id` UUID, `name` length) — endpoint exists and validates.
- `DELETE /export-jobs/{id}` bogus → 404 `resource: export`.

Export-job create/delete lifecycle intentionally not exercised end-to-end: it would create real recurring export state and ship objects to a bucket. `list_export_jobs` (200) + the probes above are the verification bar for this PR.

### Notes

- Smoke test's custom-rule lifecycle block tolerates the 501 and reports it as "blocked (expected)" — it flips to a real create→update→delete→verify-gone check the moment Scaleway enables the backend, no test change needed.
- Write tools (alert rules, export jobs) surface API errors verbatim; the server credential currently documented with `AuditTrailReadOnly` will get `permissions_denied` on writes until a write permission set is granted — descriptions say so explicitly.
- Blind spot: `nl-ams` region behavior untested (single-region org); and `evaluation_window` format for custom rules is unverified since the endpoint 501s.

src/tools/auditTrailShared.ts
```diff
@@ -0,0 +1,206 @@
+import { z } from "zod";
+import type { Config } from "../config.js";
+import { AuditTrailApiError } from "../auditClient.js";
+import { toolError } from "../output.js";
+
+export const auditRegionSchema = z
+  .enum(["fr-par", "nl-ams"])
```
gkrost: **[P1] Accept the Warsaw Audit Trail region.** This shared validator is used by every new Audit Trail tool, so a caller cannot query or manage Audit Trail in `pl-waw` even though Scaleway documents Audit Trail as available in Paris, Amsterdam, and Warsaw. Add `pl-waw` here (and keep the tool descriptions aligned) so per-call region selection does not reject a supported deployment.

gkrost (commented):
Found one blocking regional-coverage issue in the new Audit Trail tools; inline comment includes the fix.

gkrost:
**Test-driven live against the real account.** Ran the extended \`scripts/smoke-test.mjs\` (passes clean, including all 8 read-path audit tools and the confirmed-501 custom-alert-rule path), then manually exercised everything the smoke test doesn't cover:

## Preconfigured alert rules
- Baseline: all 11 rules `disabled` on this (long-running, non-fresh) account.
- \`scaleway_audit_set_alert_rules_enabled\` (additive): enabled `ApiKeyCreated` alone, verified only it flipped to `enabled`, disabled it again, verified reverted to the original all-disabled baseline. `status` filter param also confirmed working.
- Confirmed the "unknown ID rejects the whole request" claim: `[real_id, "00000000-...-000000000000"]` → `404`, and verified the real rule's status was untouched afterward (no partial application).
- \`scaleway_audit_replace_enabled_alert_rules\` - **could not verify live.** Claude Code's own auto-mode safety classifier blocked this specific call twice, even after explicit approval in-session. It's the one write tool in this PR I couldn't actually exercise; its full-replace "disables everything not listed" semantics remain unverified beyond the request-validation behavior it shares with the additive tool (also 404s + no partial application, confirmed via the additive tool).

## Export jobs
- Full lifecycle confirmed: create → appears in `list_export_jobs` → delete → gone.
- Found something worth flagging: creation triggers an **immediate backfill**, not just a future cadence - a job against a fresh empty bucket wrote 6 days of daily log objects (`2026/07/12/logs_*.json` ... `2026/07/17/logs_*.json`) within seconds. `delete_export_job` correctly doesn't clean these up (matches its own doc), which meant the throwaway test bucket needed a raw S3 call to empty before `delete_bucket` would succeed - documented in the tool description and `docs/gotchas.md` (pushed as \`66e0f16\`).

## Custom alert rules
- Re-confirmed \`list\`/\`create\` both still 501 "unknown method" - matches this PR's own caveat, not a permissions issue (I had only `AuditTrailReadOnly` on first attempt, then granted `AuditTrailFullAccess` and got the same 501 either way).

## One thing worth knowing: I modified a live credential to test this
The server's own operational credential (\`scaleway-ops-mcp_iam_and_storage\`) only had `AuditTrailReadOnly`, so every write-path tool 403'd initially. With your go-ahead I upgraded that policy's rule from `AuditTrailReadOnly` to `AuditTrailFullAccess` (kept, not reverted, per your instruction) so the write paths could actually be exercised rather than just confirming 403s. Worth knowing this server's own credential now has full Audit Trail write access, not just read.

Everything else created during testing (2 throwaway apps, 2 keys, 2 buckets, 1 export job + its 6 backfilled objects, 1 brief alert-rule toggle) was cleaned up - final app/bucket/policy counts match pre-test baselines exactly.

Pushed \`551cbbf\`/rebased to \`66e0f16\` (on top of your `pl-waw` region fix) with the export-job and alert-rule findings.

PR #12, opened 2026-08-18
feat: implement object-level operations (issue #8)

gkrost:
Closes #8.

## What
Full object-level CRUD for Object Storage, reusing `s3Client.ts`'s existing SigV4 path:

- `scaleway_s3_put_object`, `scaleway_s3_get_object`, `scaleway_s3_list_objects`, `scaleway_s3_head_object`, `scaleway_s3_copy_object`, `scaleway_s3_delete_object`, `scaleway_s3_delete_objects` (all P0)
- `scaleway_s3_get_object_tags`/`put_object_tags`, `scaleway_s3_generate_presigned_url` (both P1 - implemented rather than deferred, small extra surface given the existing client)

Binary payloads travel as base64 (`encoding: "base64"`); decoded size is capped by a new `MAX_PUT_OBJECT_BYTES` config var (default 5 MB) - multipart/large-file upload stays out of scope. `get_object` auto-detects UTF-8 text vs. binary.

## Verification (AC 2 - live via MCP only)
`scripts/smoke-test.mjs` extended with the full lifecycle against the real account:
create bucket → put text+binary → head → list → get (byte-identical round-trip, both encodings) → tags set/read → copy → presigned-URL fetch (real signature verified by fetching it, not just checking URL shape) → batch delete → delete object → bucket gone. Full run output attached below.

## Verification (AC 3 - console cross-validated, both directions)
Same session, same throwaway bucket:
- Created via MCP (`put_object`) → confirmed visible in the console Files tab (screenshot).
- Uploaded a file via the console Files tab UI → confirmed it appeared via `scaleway_s3_list_objects` with matching size/etag.
- Deleted both objects + the bucket via MCP → confirmed gone from the console's bucket list (buckets 7 → 6).

One console-only quirk documented in `docs/gotchas.md`: the console interposes a "Storage class" modal before every upload, even a single small file - `put_object` has no equivalent step.

## Docs
- `README.md` tool list + Scope + Configuration (`MAX_PUT_OBJECT_BYTES`)
- `docs/capability-gap-analysis.md` Object Storage table + Bottom line - object ops no longer listed as out of scope/wontfix
- `docs/gotchas.md` - console storage-class-modal quirk

## Test plan
- [x] `npm run build` - clean
- [x] `node scripts/smoke-test.mjs` - full lifecycle passes end to end (applications, API keys, buckets, objects, audit trail)
- [x] Console cross-validation both directions, screenshots taken
[...]

Issue #8, opened 2026-08-18
Implement object-level operations: full bucket + object CRUD via MCP, live-verified and console-cross-validated


gkrost:
# Implement object-level operations: full bucket + object CRUD, live-verified via MCP and cross-validated against the Console UI

**Direction change 2026-08-18:** this issue was previously pre-labeled `wontfix` ("this server never touches object data - only access control"). Direction changed: object operations are now in scope and this issue is the implementation tracker. The `wontfix` label is removed.

## Why

Today this server manages buckets and access control (Bucket Policies) but cannot touch the objects inside them. Any task involving actual data still drops out of MCP to the console or a raw `@aws-sdk/client-s3` script - exactly the friction this server exists to remove (the same failure mode that motivated adding the bucket lifecycle tools on 2026-08-18).

## Scope

**P0 - object operations** (S3-compatible API, reusing the existing SigV4 path in `s3Client.ts`):

| Tool | Notes |
|---|---|
| `scaleway_s3_put_object` | Upload; text or base64 binary payload; single-part with a documented size guard (multipart/large-file out of scope) |
| `scaleway_s3_get_object` | Download; returns object content; truncation guard via `MAX_OUTPUT_CHARS`; metadata alongside content |
| `scaleway_s3_list_objects` | prefix / delimiter / limit; keys + size + last-modified |
| `scaleway_s3_head_object` | Metadata only: content-type, size, etag, last-modified |
| `scaleway_s3_copy_object` | Same-bucket and cross-bucket copy |
| `scaleway_s3_delete_object` | confirm-guarded |
| `scaleway_s3_delete_objects` | Batch delete, confirm-guarded, no recursive bucket wipe |

**P1 - implemented or explicitly deferred with rationale in this issue before close:**
- Object tags: get / put
- Presigned URLs: generate (time-limited GET/PUT)

**Re-verify as part of this issue (already exist, never UI-cross-validated):**
- `scaleway_s3_create_bucket` / `list_buckets` / `delete_bucket`

**Still out of scope** (unchanged, tracked in #7): versioning, object-lock/retention, SSE, ACL/visibility, lifecycle rules, CORS, static-website config, bucket metrics/access logs, multipart upload.

## Acceptance criteria

[...]

PR #13, opened 2026-08-18
feat(s3): bucket configuration tools - visibility, versioning, object lock, website, lifecycle, tags, CORS, encryption

### LINKED ISSUE #7: Implement bucket configuration: visibility, versioning, object lock, website, lifecycle, tags, CORS, access logs (agentic spec)
# Implement bucket configuration: visibility, versioning, object lock, website, lifecycle, tags, CORS, access logs

**Direction change 2026-08-18:** this issue previously recorded these capabilities as a deliberate out-of-scope boundary ("use the console or `scw` CLI"). Direction changed: bucket configuration is now in scope and this issue is the self-contained implementation spec. Written to be executable end-to-end by an autonomous agent without further clarification - every assumption that could not be pre-verified is called out as a **verify-first** step with a concrete resolution recipe instead of being silently assumed.

## Objective

Extend the Object Storage toolset (currently only `scaleway_s3_create_bucket` / `list_buckets` / `delete_bucket` in `src/tools/buckets.ts`) with bucket-level configuration management via Scaleway's S3-compatible API, implemented, live-verified, and console-cross-validated per the protocol below.

| Thing | Where |
|---|---|
| Existing bucket tools + the now-stale "out of scope" doc comment to update | `src/tools/buckets.ts` (comment block above `registerBuckets`, plus `create_bucket`'s description referencing the console/CLI) |
| Per-region S3 client factory (SigV4 auth, `s3.<region>.scw.cloud` endpoints) + `handleS3` error mapping | `src/s3Client.ts` |
| Tool conventions: zod input schemas, `confirm: z.literal(true)` guard on destructive ops, `toolJsonResult(data, config.MAX_OUTPUT_CHARS)`, annotations `{ readOnlyHint, destructiveHint, idempotentHint, openWorldHint: true }` | any file in `src/tools/` (e.g. `apiKeys.ts` for confirm-guard style) |
| Full-replace PUT precedent (model for lifecycle/tagging/website full-replace semantics) | `scaleway_iam_set_policy_rules` in `src/tools/policies.ts` |
| Live smoke harness to extend | `scripts/smoke-test.mjs` |
| SDK already a dependency (no new deps allowed without justification) | `@aws-sdk/client-s3` in `package.json` |
| Docs to update at the end | `README.md` (Tools + Scope), `docs/capability-gap-analysis.md` (Object Storage table + Bottom line), `docs/gotchas.md` (new Scaleway-specific findings) |

Build gate: `npm run build` must pass; there is no separate lint script. Verification runs `node scripts/smoke-test.mjs` (live credential from `.env`).

## Work items

Risk classes → guards: **[R0]** read-only, no guard · **[R1]** reversible write, no confirm · **[R2]** reversible write that weakens protection or publishes content, `confirm: true` · **[R3]** irreversible/destructive, `confirm: true` + explicit warning in description. S3 semantics note: `Put*Configuration`/tagging/CORS/website calls are **full-replace** of that configuration - the request body IS the complete desired state (same contract as `set_policy_rules`); descriptions must say so.

### P0 — low-risk config (implement first)

| # | Capability | S3 API (SDK commands) | Tool(s) | Risk |
|---|---|---|---|---|
| 1 | Bucket tags | `GetBucketTaggingCommand`, `PutBucketTaggingCommand` (full-replace), `DeleteBucketTaggingCommand` | `scaleway_s3_get/put/delete_bucket_tagging` | R0 / R1 / R2(delete) |
| 2 | CORS | `Get/Put/DeleteBucketCorsCommand` (put = full-replace) | `scaleway_s3_get/put/delete_bucket_cors` | R0 / R1 / R1 |
| 3 | Versioning | `GetBucketVersioningCommand`, `PutBucketVersioningCommand` (`Status: Enabled`/`Suspended`) | `scaleway_s3_get_bucket_versioning`, `scaleway_s3_set_bucket_versioning` | R0 / R1(enable) + R2(suspend) |
| 4 | Static website | `Get/Put/DeleteBucketWebsiteCommand` (put = full-replace) | `scaleway_s3_get/put/delete_bucket_website` | R0 / R2(put: publishes an HTTP endpoint) / R1(delete) |

### P1 — security-sensitive config

| # | Capability | S3 API | Tool(s) | Risk |
|---|---|---|---|---|
| 5 | Visibility (public/private) | `PutBucketAclCommand` with canned ACL (`private` / `public-read`) + `GetBucketAclCommand` | `scaleway_s3_get_bucket_visibility`, `scaleway_s3_set_bucket_visibility` | R0 / R2(make public) + R1(make private). Making a bucket public is the highest-impact reversible op here - description must warn that public-read exposes ALL object content to the anonymous internet |
| 6 | Lifecycle rules (expiration/transition) | `Get/Put/DeleteBucketLifecycleConfigurationCommand` (put = full-replace) | `scaleway_s3_get/put/delete_bucket_lifecycle` | R0 / R3(put: rules with `Expiration` permanently delete objects once applied - confirm regardless of rule content, simpler and safer than conditional guards) / R3(delete: removes protection the caller may rely on) |
| 7 | Access logs (delivery to target bucket) | `Get/Put/DeleteBucketLoggingCommand` (put = full-replace; `TargetBucket` must exist - same constraint pattern as `create_export_job` in `auditTrailExports.ts`) | `scaleway_s3_get/put/delete_bucket_logging` | R0 / R1 / R1 |

### P2 — investigate-first (resolve hypothesis before writing any code)

| # | Capability | Hypothesis to verify | Resolution recipe if confirmed | If refuted |
|---|---|---|---|---|
| 8 | Object lock + retention mode | Enable-once semantics: like AWS, object lock can only be enabled at bucket **creation** (Scaleway console shows an "Object Lock" checkbox on create) and can **never** be disabled; per-object retention/modes are object-level ops (already #8-object territory). `PutObjectLockConfigurationCommand` on an existing bucket may therefore be rejected by Scaleway even though the SDK ships it | Implement `create_bucket` opt-in param `object_lock: true` (R3 confirm at creation via that param's description) + `scaleway_s3_get_object_lock_configuration` (R0). Do NOT offer enable-on-existing-bucket unless live probe proves Scaleway allows it; if it allows it, gate behind R3 confirm | Implement as normal R3 toolset: get/put (enable-only, confirm) |
| 9 | SSE encryption | Scaleway encrypts all Object Storage at rest (managed SSE) with **no per-bucket toggle** - the console Settings tab and S3 API expose no encryption config | No tool. Close the item with a short doc note (README/gap-analysis: "SSE: always-on platform-side, nothing to configure") | Implement get/put per actual API surface found |
| 10 | Metrics | Bucket metrics are a Cockpit/console feature, not an S3 API surface | No tool; doc note pointing to Cockpit | Implement whatever read surface actually exists |

**Verification-before-truth rule for P2 (and any claim about Scaleway behavior):** every hypothesis must be resolved against (a) Scaleway's S3-compatibility docs (https://www.scaleway.com/en/docs/storage/object/api-reference/object-storage-api-coverage/) and/or (b) a live stateless probe (create throwaway bucket → attempt the operation → read the exact error/response → delete bucket). Record the exact response text in the PR description. Unverified claims must be labeled as such - never reported as fact.

## Implementation requirements

1. **Conventions:** follow `src/tools/` exactly - zod schemas with per-field `.describe()` guidance for the agent caller, confirm-guards per the risk table, `handleS3` for error mapping, `toolJsonResult` + `MAX_OUTPUT_CHARS` truncation, correct annotations per risk class (R0 → `readOnlyHint: true, idempotentHint: true`; R1/R2 → `readOnlyHint: false, destructiveHint: false, idempotentHint: true` where genuinely idempotent; R3 → `destructiveHint: true`).
2. **File layout:** new tools in `src/tools/bucketConfig.ts` (one file is fine at this size; split only if it exceeds ~500 lines), registered in `src/index.ts` next to `registerBuckets`. Do not touch `buckets.ts` behavior beyond updating its stale scope comment and `create_bucket` description.
3. **Full-replace UX:** for lifecycle/tagging/website/logging puts, the description must state "the request replaces the complete configuration - omitted elements are removed" (mirroring `set_policy_rules` precedent).
4. **Cleanup ordering:** configuration attached to a bucket may block `delete_bucket` (e.g. logging target references, non-empty versioned objects). Smoke-test cleanup must tear down in reverse order (delete configs → delete objects/versions → delete bucket) and must clean up **even on mid-script failure** (try/finally around each throwaway resource).

## Verification protocol (both directions, per feature)

For every P0/P1 feature AND the re-verified `create/list/delete_bucket`: make the change via MCP tool call, confirm in the Scaleway console (Object Storage → bucket → **Settings** tab / Files tab) in the same session, then where a reverse direction exists (console → MCP), exercise it once (e.g. enable versioning in console → `get_bucket_versioning` reports `Enabled`). Record a one-line evidence row per feature in the PR description:

| Feature | MCP evidence (tool + key response field) | Console evidence (tab + observed state) | Reverse-direction check |
|---|---|---|---|
| tags | `put_bucket_tagging` → `get_bucket_tagging` echoes set | Settings tab shows tags | edit tag in console → get reflects it |
| CORS / website / lifecycle / logging / visibility / versioning / object lock | same pattern | respective Settings sections | where mutable in console, one round-trip |

Constraints: single throwaway bucket (`mcp-smoke-test-config-<ts>`) reused for all features, torn down at the end; `nl-ams` explicitly untested (single-region org) - state that in the PR; object-lock enable is permanent for that bucket, so the throwaway bucket with object lock enabled must simply be deleted (empty + lock enabled deletable? **verify-first** sub-case: if delete of a lock-enabled empty bucket is refused, leave it deleted-able after removing all versions, else flag the residue in the PR rather than forcing it).

## Acceptance criteria

1. Every P0 and P1 row exists as implemented tools per the risk table, or carries a written justification in the PR for deviation (with the live probe evidence that motivated it).
2. Every P2 hypothesis is resolved with recorded evidence and either implemented per the recipe or closed as doc-only with the doc updated.
3. `npm run build` green; `scripts/smoke-test.mjs` extended with a config-lifecycle section that passes end-to-end against the live account and leaves **zero residue** (or documents forced residue per the object-lock sub-case).
4. Cross-validation matrix (above) filled in the PR description with real evidence, both directions.
5. Docs updated: README tool list + Scope section; `docs/capability-gap-analysis.md` Object Storage table rows for items 1-10 and Bottom line; `docs/gotchas.md` gains any new Scaleway-specific behavior discovered (especially P2 findings); stale scope comment in `src/tools/buckets.ts` and `create_bucket` description corrected.
6. PR references this issue (`Closes #7`), one commit series following repo style (`feat(s3): ...`).

## Explicitly out of scope

- Object-data operations (put/get/delete objects, metadata, presigned URLs) → #8
- Multipart upload, bucket policy tools (already exist), anything listed as still out of scope in the gap analysis

gkrost:
Closes #7

## What

21 new tools in \src/tools/bucketConfig.ts\, risk-classed per the issue spec (R0 reads unguarded; R1 reversible writes; R2/R3 confirm-guarded):

- **Tags**: \get/put/delete_bucket_tagging\ (put = full-replace)
- **CORS**: \get/put/delete_bucket_cors\ (put = full-replace)
- **Versioning**: \get_bucket_versioning\, \set_bucket_versioning\ (suspend requires confirm)
- **Website**: \get/put/delete_bucket_website\ (put requires confirm - publishes endpoint)
- **Visibility**: \get/set_bucket_visibility\ (public requires confirm; AllUsers-grant detection)
- **Lifecycle**: \get/put/delete_bucket_lifecycle\ (put/delete confirm-guarded - expiration permanently deletes)
- **Encryption**: \get/put/delete_bucket_encryption\ (declarative only - see findings)
- **Object Lock**: \get_object_lock\, \nable_object_lock\ (confirm + one-way warnings, versioning prerequisite handled in-tool)
- **Logging / metrics: no tools** - Scaleway S3 API returns \NotImplemented\ (finding F2)

## P2 verify-first findings (live-probed 2026-08-18, fr-par, throwaway buckets)

| # | Hypothesis | Verdict | Evidence |
|---|---|---|---|
| F1 | Object lock enable-once-at-create only | **Partially refuted** | Enable on existing bucket works BUT only after versioning Enabled (\InvalidBucketState\ otherwise). Disable is impossible at XML-schema level (\MalformedXML\ for \Disabled\). Versioning frozen after lock (\InvalidBucketState\). \CreateBucket\ with \ObjectLockEnabledForBucket\ is **silently ignored** (200 OK, no lock, no versioning). Empty locked bucket deletes fine. |
| F2 | Logging supported | **Refuted** | \Put/GetBucketLogging\ -> \NotImplemented: Action not implemented\ |
| F3 | Metrics = Cockpit-only | **Confirmed** | \GetBucketMetricsConfiguration\ -> \NotImplemented\ |
| F4 | SSE always-on, no API | **Refuted** | \Get/Put/DeleteBucketEncryption\ all work - but config is **declarative**: fresh bucket returns 200 with empty config (AWS errors instead); put/delete change nothing about actual at-rest encryption (always-on platform-side). Tools implemented with descriptions saying exactly this. |

All findings also recorded in \docs/gotchas.md\.

## Verification evidence matrix

**MCP side: fully verified live** (\scripts/smoke-test.mjs\, throwaway bucket, zero residue - teardown verified by list_buckets):

| Feature | MCP evidence (tool + verified outcome) | Guard negative-test |
|---|---|---|
| tags | put(2) -> get echoed 2/2 -> delete -> get errors NoSuchTagSet | delete confirm |
| CORS | put(1) -> get echoed origins/methods/max_age -> delete | - |
| versioning | Enabled -> get=Enabled -> Suspended -> get=Suspended -> re-Enabled | suspend-without-confirm rejected |
| website | put(index+error) -> get echoed both -> delete | put confirm |
| visibility | initial private -> public (AllUsers grant observed) -> private | public-without-confirm rejected |
| lifecycle | put(1 rule, expire 1d) -> get echoed -> delete | put+delete confirm |
| encryption | put AES256 -> get echoed -> delete | - |
| object lock | instructive error w/o versioning -> enable (auto-enables versioning) -> get=Enabled -> **versioning suspend blocked while locked** | lock-without-versioning instructive error |
| bucket lifecycle | re-verified create/list/delete + config bucket teardown zero-residue | delete confirm |

**Console UI side: NOT verified - requires a human with console access.** The agent has API credentials only (no console login); the reverse-direction checks (console edit -> MCP get) equally pending. The API-side evidence above is the complete automated verification; console cross-validation rows from the issue remain open for a reviewer with console access - flagging rather than claiming.

## Deviations from the issue spec

- \delete_bucket_tagging\ implemented as confirm-guarded per issue risk table (R2) though impact is low.
- Object lock recipe differs from the issue's two hypotheses: not create-time-only, not enable-on-existing-always-works - it's enable-after-versioning, implemented accordingly with \nable_versioning_if_needed\.
- \
l-ams\ untested (single-region org).

Docs updated: README (tools + scope), \docs/capability-gap-analysis.md\ (Object Storage table, bottom line), \docs/gotchas.md\ (4 new entries), stale scope comments in \uckets.ts\ corrected. \
pm run build\ green.


src/tools/bucketConfig.ts
```diff
@@ -0,0 +1,528 @@
...
+      }),
+  );
+
+  // ---------- Encryption (declarative config) ----------
+  server.registerTool(
+    "scaleway_s3_get_bucket_encryption",
+    {
+      title: "Get Scaleway bucket encryption config",
+      description:
+        "Read the bucket's S3 default-encryption configuration. IMPORTANT (live-verified 2026-08-18): Scaleway returns an EMPTY config (not an error) for buckets where none was ever set, and Scaleway encrypts all Object Storage at rest platform-side regardless - this config is declarative S3-API-compatibility metadata, not an on/off switch for actual encryption.",
```
gkrost: **Live-tested, and this claim needs revising.** I created a throwaway bucket, checked its console Settings tab: "Encryption type: Disabled". Then called `scaleway_s3_put_bucket_encryption` (AES256) via MCP and reloaded the console page: "Encryption type: SSE-ONE encryption with Scaleway Object Native Encryption keys". Called `scaleway_s3_delete_bucket_encryption` and reloaded again: back to "Disabled". This is a real, reproducible, console-visible toggle between a named encryption mode and "Disabled" - not an inert compatibility label. It directly contradicts "this config is declarative S3-API-compatibility metadata, not an on/off switch for actual encryption" and the accompanying claims on `put_bucket_encryption` (line ~436, "Scaleway encrypts at rest regardless") and `delete_bucket_encryption` (line ~461, "Data stays encrypted at rest ... this only removes the S3-API config object"). It's possible there's a baseline platform encryption layer underneath that's genuinely always-on regardless of this setting (which would make the *DeleteBucketEncryption never actually decrypts your data* framing directionally correct), but the tool descriptions currently assert as fact that this setting has no effect at all - and the console evidence says otherwise. This is exactly the kind of claim the issue's "verification-before-truth" rule was written for, and the PR explicitly could not check the console side ("agent has API credentials only"). Recommend re-verifying against Scaleway's actual encryption docs and softening/correcting these three descriptions (and the matching docs/gotchas.md entry at line 236) before merge - this is user-facing safety-relevant text, not a cosmetic issue.

docs/gotchas.md
```diff
@@ -222,3 +222,21 @@ key) that turned out to be a permission classifier gating those specific actions
...
+- `PutObjectLockConfiguration` on a bucket without versioning fails `InvalidBucketState: Versioning must be 'Enabled'...` - so `scaleway_s3_enable_object_lock` checks/enables versioning itself (`enable_versioning_if_needed`).
+- Disabling is impossible at the protocol level: `ObjectLockEnabled: "Disabled"` is rejected `MalformedXML` (the schema has no disabled state) - a true one-way door, documented as such in the tool.
+- After lock is enabled, versioning is frozen: `PutBucketVersioning Suspended` fails `InvalidBucketState: An Object Lock configuration is present on this bucket, so the versioning state cannot be changed.`
+- An empty bucket with lock + versioning enabled still deletes normally (`DeleteBucket` OK) - no residue trap there.
+
+## S3 bucket encryption config on Scaleway is declarative metadata, not an encryption switch (issue #7)
```
gkrost: Same issue as the inline comment on `scaleway_s3_get_bucket_encryption` in bucketConfig.ts: live console cross-check shows PutBucketEncryption/DeleteBucketEncryption toggling the bucket's console-displayed "Encryption type" between "SSE-ONE encryption with Scaleway Object Native Encryption keys" and "Disabled" - this entry's "declarative metadata, not an encryption switch" conclusion should be re-verified and corrected/qualified.

src/tools/bucketConfig.ts
```diff
@@ -0,0 +1,528 @@
...
+        );
+        return toolJsonResult({ bucket, applied: tags.length, tags }, config.MAX_OUTPUT_CHARS);
+      }),
+  );
+
+  server.registerTool(
+    "scaleway_s3_delete_bucket_tagging",
+    {
+      title: "Delete Scaleway bucket tags",
+      description: "Remove ALL tags from a bucket. Requires confirm=true. Reversible in the sense that tags can be re-applied, but the current set is lost.",
```
gkrost, line 86: Nit, non-blocking: `idempotentHint: false` here looks off - repeated `DeleteBucketTagging` calls on an already-untagged bucket succeed (204) rather than erroring, live-verified during this review, so this looks idempotent in practice. Doesn't affect correctness of the tool itself, just the advertised annotation.

gkrost (commented):
Live-tested this against the real account (fresh throwaway buckets, zero residue verified via list_buckets before/after) and cross-checked against the console UI, which the PR description flagged as unverified. Full summary as a separate comment below. One finding worth flagging as blocking before merge: the encryption-config claim ("declarative only, changes nothing about actual at-rest encryption") does not hold up against what the console shows - see the inline comment on `scaleway_s3_get_bucket_encryption`.

gkrost:
## Review summary (live-tested against the real account)

**Setup:** Checked out at `0c37e69`, `npm run build` clean, `node scripts/smoke-test.mjs` passed end-to-end including the full bucket-config lifecycle section. Baseline `list_buckets` recorded before testing (6 pre-existing buckets); re-checked after - identical, zero residue. All testing used freshly created `mcp-pr13-review-*` / `mcp-smoke-test-*` throwaway buckets, all deleted by the end of each run.

**Code review:** Confirm-guards match the R0-R3 risk table from issue #7 exactly, including the one documented deviation (`delete_bucket_tagging` guarded despite low impact, per the PR's own note). Full-replace semantics are documented in every relevant tool description. `handleS3`/`toolJsonResult`/`toolError` used consistently, no ad-hoc error handling. `buckets.ts` diff is a clean doc-only update (stale scope comment + `create_bucket` description), no behavior change.

**Independently reproduced live (beyond the smoke test):**
- Object lock: enable requires versioning first (instructive error otherwise), auto-enables versioning when asked, freezes versioning at Enabled, and calling enable twice is safely idempotent. An empty locked+versioned bucket deletes cleanly - no orphaned residue.
- `CreateBucket` create-time object-lock flag is not exposed by this PR's tools at all (correct, since the PR found it's silently ignored by Scaleway).
- Bucket logging and bucket metrics: confirmed via raw SDK calls (`GetBucketLoggingCommand`, `GetBucketMetricsConfigurationCommand`) that Scaleway returns `NotImplemented` - correctly no tools were built for these.
- Lifecycle rule with `prefix` omitted works fine (no XML/Filter validation issue).
- Every confirm-guarded tool correctly rejects the call without `confirm: true` and proceeds with it.
- `set_bucket_visibility` to public: tested only on a throwaway bucket created solely for this, confirmed via the console (bucket list showed "Public", Settings tab matched), then set back to private and the bucket deleted.

**Console UI cross-check (the part the PR author explicitly could not do):** a Scaleway console tab was available and authenticated. Cross-checked tags, visibility, and encryption on a throwaway bucket - all matched, with one important exception below.

**Blocking finding - encryption description is likely wrong:** `scaleway_s3_get/put/delete_bucket_encryption` describe the encryption config as "declarative S3-API-compatibility metadata... Scaleway encrypts at rest regardless" and claim put/delete "change nothing about actual at-rest encryption." Live console check contradicts this: a fresh bucket's Settings tab showed "Encryption type: Disabled"; after `put_bucket_encryption` (AES256) it showed "Encryption type: SSE-ONE encryption with Scaleway Object Native Encryption keys"; after `delete_bucket_encryption` it went back to "Disabled." That's a real, reproducible, console-visible state change, not an inert compatibility label. Posted as an inline comment on the relevant lines (bucketConfig.ts ~417/436/461, docs/gotchas.md ~236) - recommend re-verifying against Scaleway's actual encryption docs before merge, since this is safety-relevant text an agent caller would rely on.

**Minor/non-blocking:** `delete_bucket_tagging`'s `idempotentHint: false` annotation looks wrong - repeated deletes on an untagged bucket succeeded rather than erroring when I tested it.

**Not verifiable in this session:** nothing else - browser access to the console was available and used, so console cross-validation is no longer an open gap except to the extent the encryption finding above needs a deeper dive into Scaleway's docs to fully characterize (vs. just knowing the current description is wrong).

**Verdict:** Solid implementation, conventions followed correctly, confirm-guards match spec, and almost every claim held up under live + console testing. The encryption description issue is real and should be fixed before merge since it's a factual claim baked into tool descriptions that a calling agent would trust.

gkrost:
Fixed both review findings and merge conflicts, pushed as \`192b5c2\` (merge of \`main\`, which now includes #12).

## Merge conflicts (README.md, docs/capability-gap-analysis.md, docs/gotchas.md)
Both this PR and #12 (object operations, merged since this PR opened) touched the same doc sections. Resolved by combining both sets of additions - the tool lists, scope statements, and capability tables now correctly reflect that both bucket config (#7) and object CRUD (#8) are implemented.

## Finding 1 (blocking): encryption claim was wrong - re-verified independently, not just trusted either side
Re-tested live against a fresh throwaway bucket and cross-checked the console's Settings tab directly (not relying on the review agent's account of it):

- \`put_bucket_encryption\` (AES256) → console "Encryption type" flips from \`Disabled\` to \`SSE-ONE encryption with Scaleway Object Native Encryption keys\`.
- \`delete_bucket_encryption\` → reverts it back to \`Disabled\`.

Confirmed: this is a real, functional, console-visible setting - the original claim ("declarative S3-API-compatibility metadata only, changes nothing about actual at-rest encryption") was false. Corrected in all four places it appeared: the three tool descriptions in \`src/tools/bucketConfig.ts\`, \`docs/gotchas.md\` (rewrote the entry to explain what was wrong and why, rather than just swapping the claim), \`README.md\`, and \`docs/capability-gap-analysis.md\`. I was careful not to overcorrect into an equally-unverified new claim about what "SSE-ONE" does cryptographically - the description now states only what's actually confirmed (the setting is real and toggleable), not a guess about the mechanism.

## Finding 2 (minor): \`delete_bucket_tagging\`'s \`idempotentHint\`
Live-verified: calling it twice in a row on a bucket with no tags set succeeds identically both times, no error. Changed \`idempotentHint\` from \`false\` to \`true\` to match observed behavior.

## Verification
\`npm run build\` clean, full \`scripts/smoke-test.mjs\` run passes end to end post-merge (including the complete bucket-config lifecycle: tags, CORS, versioning, website, visibility, lifecycle rules, encryption, object lock), and the encryption/idempotency fixes were independently re-verified live + via console before being written up, not assumed from the review agent's report. \`gh pr view\` now shows \`MERGEABLE\`; CI is running on the fresh push.

PR #15, opened 2026-08-18
feat(iam): human User management tools - list/get/create/update/delete, lock/unlock, password/username, MFA delete

### LINKED ISSUE #4: Implement IAM Users management: list/get/create/update/delete, lock/unlock, password/username, MFA (agentic spec)
# Implement IAM Users management (agentic spec)

**Scope change 2026-08-18:** previously closed as out of scope ("this server manages Applications, i.e. non-human identities, only"). Direction changed: human-User management is now in scope. This server's policy/rule tools already accept a `user_id` principal — after this issue they can actually manage the whole lifecycle of those users.

## Objective

Extend the IAM toolset with human-User management: list/get/create(invite)/update/delete, lock/unlock, password/username changes, MFA OTP lifecycle, per the API surface below.

## API surface (verified against https://www.scaleway.com/en/developers/api/iam/users/ , 2026-08-18)

All `https://api.scaleway.com/iam/v1alpha1/...`, same auth path as `src/iamClient.ts` (`iamRequest`).

| Endpoint | Notes |
|---|---|
| `GET /users` | list; filters incl. user_id, name, email |
| `POST /users` | create — but see hypothesis H1 (invitation semantics) |
| `GET /users/{user_id}` | |
| `PATCH /users/{user_id}` | name/email/tags/protected |
| `DELETE /users/{user_id}` | named "Delete a **guest** user" — see H2 |
| `POST /users/{user_id}/lock` | |
| `POST /users/{user_id}/unlock` | |
| `POST /users/{user_id}/update-password` | see H3 — admin resets another human's password |
| `POST /users/{user_id}/update-username` | login identity change |
| `POST /users/{user_id}/mfa-otp` (create), `POST .../mfa-otp/validate`, `DELETE .../mfa-otp` | enrollment flow; validate requires the OTP code from the user's authenticator |
| `GET /users/{user_id}/grace-periods` | read-only extra, cheap to add |
| `users/{id}/connections` (get/initiate/join/remove) | SSO session management — **out of scope** unless trivial |

Scaleway's own IAM docs state (Technical Limitations, checked 2026-08-18): *"IAM users cannot be created within Scaleway Organizations, they can only be invited to join them"* — while a POST /users endpoint exists. This tension is H1 below and shapes the whole tool design.

| Thing | Where |
|---|---|
| IAM client (`iamRequest`, `iamListAll`, `withIamErrorHandling`) | `src/iamClient.ts` |
| CRUD + confirm-guard template (`create/update/list/delete` quartet) | `src/tools/apiKeys.ts`, `applications.ts` |
| Registration point | `src/index.ts` |
| Where user_id principals already appear | `create_policy`/`set_policy_rules` in `policies.ts` (docs should cross-link after this lands) |
| Permission-set catalog for the credential's own needs | `scaleway_iam_list_permission_sets` |
| Smoke harness | `scripts/smoke-test.mjs` |
| Docs to update | `README.md`, `docs/capability-gap-analysis.md` (IAM table "Users" row), `docs/gotchas.md` |

Build gate: `npm run build`; verification `node scripts/smoke-test.mjs` (live `.env` credential).

## Work items

Risk classes (as in #7): **[R0]** read · **[R1]** reversible write · **[R2]** access-affecting write, `confirm: true` · **[R3]** identity-affecting / lockout-capable, `confirm: true` + explicit warning. Users are HUMANS: every mutating tool's description must make the human impact explicit (unlike Applications, a mistake locks a person out, not a service).

| # | Tool | Op | Risk | Notes |
|---|---|---|---|---|
| 1 | `scaleway_iam_list_users` | list (name/email filters) | R0 | |
| 2 | `scaleway_iam_get_user` | get | R0 | include status fields (locked/protected/mfa flags as returned) |
| 3 | `scaleway_iam_create_user` | create | R2 | confirm: sends a REAL invitation email to the given address — never call speculatively. Fields per API (email, name, tags?). Behavior per H1. |
| 4 | `scaleway_iam_update_user` | patch (name/tags; email if allowed) | R1 | |
| 5 | `scaleway_iam_delete_user` | delete | R3 | confirm + warning. Guest-only per API name (H2): tool must surface the API's error verbatim when the target is a Member/Owner. MUST refuse to delete the org Owner (probe H2; belt-and-braces: refuse when user_id == the org's owner id, discoverable via GET /users or org info). |
| 6 | `scaleway_iam_lock_user` | lock | R2 | confirm + warning: blocks the person's console/API login immediately. Expected use: offboarding compromise response. |
| 7 | `scaleway_iam_unlock_user` | unlock | R1 | |
| 8 | `scaleway_iam_update_user_password` | admin password reset | R3 | confirm + warning: resetting another human's password is account-takeover-grade power. Response may contain a generated password or require delivery semantics — treat any secret like `create_api_key` does (returned once, caller must capture; NEVER log). See H3. |
| 9 | `scaleway_iam_update_user_username` | login change | R2 | confirm: changes what the human types to log in |
| 10 | `scaleway_iam_list_grace_periods` | read | R0 | cheap completeness; skip if awkward |
| 11 | MFA OTP create/delete/validate | enrollment + removal | R3 | confirm + warning: DELETING a user's MFA factor weakens their account security immediately; creating+validating an OTP requires the user's authenticator code, so validate is only completable with the human's cooperation — design validate to take the code, and document that create without the human completing enrollment may leave MFA half-configured (probe H4 for exact flow states). If probing shows the flow is self-user-only (403 for admins), implement list-only or skip, per evidence. |
| 12 | connections endpoints | — | — | out of scope unless probe shows trivial value |

## Verify-first hypotheses (probe before coding the affected tools; record exact responses in the PR)

- **H1 create-vs-invite:** does `POST /users` (a) create a usable Member and email them a set-password link, (b) send an invitation the recipient must accept, or (c) reject outright on this org type? Docs' "invite-only" limitation vs the endpoint's existence must be resolved empirically. Probe: create with a **disposable mailbox the operator controls** (ask the user of this issue for one before running — do NOT use a fabricated address that would email a stranger). Record the response shape and whether `status` is pending/active.
- **H2 delete semantics:** delete a (i) guest vs (ii) member vs (iii) org owner → expect (i) 200, (ii)+(iii) instructive 4xx. Residue rule: any user created during probing MUST be deletable before smoke test ends; if a created Member proves undeletable, document the residue in the PR and leave the account locked instead (explicitly stated, not silent).
- **H3 password reset flow:** does update-password return a plaintext/one-time secret, require `password_reset_token`, or email the user directly? Determines whether the tool's response carries a secret (handle like API-key secrets: never logged, returned once).
- **H4 MFA flow + permissions:** can this credential (an Application's key) manage MFA OTPs for another user at all, or self-scoped only? Are there intermediate states (created-but-unvalidated)?
- **H5 lock semantics:** does lock invalidate active sessions/tokens immediately? Is unlock immediate? What does a locked user's `GET` show?
- **H6 permission sets:** which permission set names gate these endpoints (IAMUserManager? IAMManager?) — enumerate via `scaleway_iam_list_permission_sets`; extend the server credential's policy via `scaleway_iam_set_policy_rules` **only with the user's explicit go-ahead** (production credential state).

## Verification protocol

1. **Disposable-mailbox user (only if the user provides one):** full lifecycle live — create → get echoes → list contains → update tags/name → lock (get shows locked) → unlock → delete → verify gone. This is the only path that exercises #3/#5 for real.
2. **No disposable mailbox available:** create/delete verified by negative tests only (invalid email format, bogus user_id → instructive errors surfaced verbatim); lock/unlock/update NOT run against real humans — say so plainly in the PR. List/get verified live read-only against the org's real users (output redacted to ids/status, no emails printed into logs beyond what the tools return).
3. **Guards:** every R2/R3 tool's missing-confirm rejection asserted in the smoke test (pattern: `expectError` blocks from the bucket-config section).
4. **Zero residue or explicitly documented forced residue** (per H2 rule).
5. Console cross-validation (read direction): console IAM > Users list matches `list_users` ids/status; note honestly what was checked. Mutation-direction cross-validation only if a disposable mailbox was used.

## Acceptance criteria

1. Tools #1-#10 (+#11 if H4 permits) implemented per the risk table; deviations justified in the PR with probe evidence.
2. H1-H6 resolved with recorded evidence; unverified behavior labeled as such, never presented as fact.
3. `npm run build` green; smoke test extended per protocol (path 1 if mailbox provided, else path 2), passing, zero undocumented residue.
4. README (tools + scope: no longer "Applications only"), gap-analysis Users row, gotchas entries for H1-H5 findings.
5. PR references this issue (`Closes #4`), includes the live-vs-negative-test evidence matrix.

## Out of scope

Groups (#5), SSH keys/JWTs/quotas/SAML/SCIM (#6), connections endpoints (unless trivial), anything touching the org Owner's account.

gkrost:
Closes #4

## What

11 tools in `src/tools/users.ts`, human-impact risk classes per the issue spec (every mutating description states its effect on the person):

| Tool | Guard | Notes |
|---|---|---|
| `list_users` / `get_user` | - | compact view: id, email, type, status, locked, mfa_enabled |
| `create_user` | confirm | sends a REAL invitation email; semantics labeled unverified (H1) |
| `update_user` | - | profile fields only |
| `delete_user` | confirm + owner-refused | tool-side refusal for `type === 'owner'`; guest-scoped API errors surfaced verbatim |
| `lock_user` / `unlock_user` | lock: confirm | lock = compromise-response, immediate |
| `update_user_password` | confirm | admin-SET password (API takes it explicitly, 1-72 bytes) - tool never returns a password |
| `update_user_username` | confirm | changes the person's login identity |
| `delete_user_mfa_otp` | confirm | security-weakening; enrollment deliberately not a tool (needs the human's authenticator code) |
| `list_user_grace_periods` | - | read-only completeness |

## Hypotheses resolved (all evidence from live probes, 2026-08-18)

| H | Verdict | Evidence |
|---|---|---|
| H1 create-vs-invite | **Partially unresolved** | `POST /users` exists; missing-type probe -> 400 `type: value is required`; email validation fires FIRST so the type enum can't be probed without emailing a real address. No disposable mailbox -> `type: "guest"` default inferred from guest-scoped delete endpoint, description labeled UNVERIFIED. |
| H2 delete semantics | Shape verified | all mutations -> clean `404 resource: user` on bogus ids. Owner happy-path unprobed (owner refusal is tool-side). |
| H3 password flow | **Resolved** | admin-set: explicit `password` field, constraint `1-72 bytes` (empty-body probe). NOT a generated-secret response. |
| H4 MFA flow | Scoped out | enrollment needs the human's authenticator code; only delete implemented, per spec fallback |
| H5 lock semantics | Unverified (flagged) | no real human to lock; bogus-id 404 shape verified only |
| H6 permissions | **Resolved** | `IAMUserManager`/`IAMUserReadOnly` gate the API; credential 403'd before grant |

## Operational note (user-approved)

The server credential's policy was extended with `IAMUserManager` (org scope, added to the existing IAM rule) **via this server's own `set_policy_rules`** - full-replace done preserving all existing rules (before/after verified). Probed first, user go-ahead asked and given per the issue's guardrail.

## Verification evidence (path 2 - no disposable mailbox)

- **Live reads:** `list_users` -> 1 user (org owner, `type: owner`, mfa true); `get_user` echoes.
- **Guards (6/6):** missing-confirm rejected for create/delete/lock/password/username/mfa-delete; owner deletion refused tool-side with the OWNER message.
- **API error shapes (7/7):** bogus-id mutations -> 404 `resource: user` surfaced verbatim; grace-periods -> 404.
- **Create negative:** invalid email rejected (zod layer) before any API call/email.
- **Zero residue:** trivially - zero writes to real humans; no throwaway resources.
- **Not verified (flagged):** mutation happy-paths on a real user (lock/unlock of the org owner was deliberately NOT attempted); create/invite email flow; console cross-validation beyond read-direction knowledge.

Docs: README (tools + scope), gap-analysis Users row + bottom line, gotchas (invite-vs-create, owner type-flag, admin-set password, permission grant). Build green, full smoke test passes.


src/tools/users.ts
```diff
@@ -0,0 +1,261 @@
...
+      inputSchema: {
+        user_id: z.string().uuid(),
+        password: z.string().min(8).max(72).describe("The NEW password (1-72 bytes API limit; >=8 enforced here). Not returned in the response."),
+        confirm: confirmField,
+      },
+      annotations: { readOnlyHint: false, destructiveHint: true, idempotentHint: true, openWorldHint: true },
+    },
+    async ({ user_id }) =>
+      withIamErrorHandling(async () => {
+        const u = await iamRequest<User>(config, "POST", `/users/${user_id}/update-password`);
```
gkrost: **Bug: the new password is never sent to the API.**

`update_user_password`'s handler destructures only `{ user_id }` and calls
`iamRequest<User>(config, "POST", `/users/${user_id}/update-password`)` with **no body** — the
`password` argument the schema requires (and that the tool description promises: "the new password
is supplied BY THE CALLER... API takes it explicitly") is silently dropped. `iamRequest`'s `body`
param is optional and omitting it sends `body: undefined` (see `iamClient.ts` — no body is
serialized at all), so this call almost certainly 400s on a missing required field, or at best does
something other than what the caller asked for. Confirmed by reading the code, not by a live call.

Corroborating: `docs/gotchas.md` in this same PR states "`update-password` takes an explicit
caller-supplied `password`" — so the correct API shape was verified during probing, it just didn't
make it into the implementation. And the smoke-test's bogus-id loop (the one that actually invokes
the handler with `confirm: true`) omits `scaleway_iam_update_user_password` — it only appears in
the earlier *missing-confirm* loop, which never reaches this code because schema validation rejects
it first. That's how this shipped untested.

Fix: destructure `password` too and pass `{ password }` as the request body.

gkrost:
**Structural: this PR is currently `CONFLICTING`/`DIRTY` against `main`, not mergeable as-is.**

`gh pr view 15 --json mergeable,mergeStateStatus` → `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`.

Root cause: `feat/iam-users-issue-4` was branched from `feat/bucket-config-issue-7` before the latter merged as **#13**. Since then #13 landed on `main` independently, so `main` and this branch each have their *own* commit that adds `src/tools/bucketConfig.ts` and edits `src/tools/buckets.ts` — same end content (I diffed both files against current `main`: byte-identical once CRLF is normalized, zero real delta), but git sees "added independently on both sides" and flags a conflict rather than silently reconciling it. That's also why `gh pr view --json files` still lists `bucketConfig.ts` as +528/-0 and `buckets.ts` as +9/-9 even though neither actually changes relative to current `main`.

Fix: rebase `feat/iam-users-issue-4` onto current `main` (`git rebase github/main`) and resolve the `bucketConfig.ts`/`buckets.ts` conflicts by taking `main`'s version — the diff should shrink to just the real Users-related files (`users.ts`, `index.ts`, `README.md`, `docs/*`, `scripts/smoke-test.mjs`). I'm not doing this rebase myself since it rewrites this branch's history and the local checkout at `C:/Users/gerno/dev/git/scaleway-ops-mcp-server` may still be mid-session on this exact branch — flagging for whoever owns that session, or happy to do it if asked.

Also see the inline comment on `src/tools/users.ts` — `scaleway_iam_update_user_password` never sends the `password` field to the API (the handler destructures only `{ user_id }`), which I'm fixing in a follow-up commit since it's a small, isolated, high-confidence bug.

src/tools/users.ts
gkrost: Verified correct — pulled your fix commit (24425b9) and confirmed by full read + live re-run: \password\ is destructured and sent as \{ password }\ (users.ts:206), the response still never echoes it, and the regression guard passes live (bogus id + valid password → 404 \
esource is not found\, proving the body reached validation-past id-lookup rather than tripping the password-constraint 400). Ordering claim also cross-checked against the original probe data (empty body → 400 password constraint; valid body + bogus id → 404) — the guard genuinely discriminates the two cases. Good catch.

gkrost:
## Full MR review (all files read) + both review findings addressed

### Review finding 1: password body omission (`src/tools/users.ts`)

**Confirmed real, fix verified.** Original handler destructured only `{ user_id }` and called `update-password` with no body — the schema-required `password` was silently dropped (would 400 on the missing field, never setting any password). The fix commit (`24425b9`, reviewer-pushed) is correct and minimal: `password` destructured and sent as `{ password }` (now users.ts:204-206). Verified three ways: full file read, `npm run build` green, live smoke re-run where the new regression guard passes — bogus id + valid password yields 404 `resource is not found` (id lookup), not the 400 password-constraint error a body-omission would produce. The guard genuinely discriminates the two failure modes (ordering cross-checked against the original probe data). Reply posted inline.

### Review finding 2: CONFLICTING/DIRTY against main

**Resolved by rebase.** Root cause as diagnosed: branch was cut from `feat/bucket-config-issue-7` before it merged as #13, so `bucketConfig.ts`/`buckets.ts` were added independently on both sides. Rebased onto `github/main` (`git rebase --skip` for the already-landed #7 commit, merge commit dropped by the rebase). PR is now **MERGEABLE**, CI **build: pass**, and the file list shrank to exactly the six real Users files: `src/tools/users.ts`, `src/index.ts`, `scripts/smoke-test.mjs`, `README.md`, `docs/capability-gap-analysis.md`, `docs/gotchas.md`.

### Full-diff review notes (non-blocking nits, no action requested)

1. **`update_user` empty-body PATCH** — if a caller passes no `tags`/`first_name`/`last_name`, the handler still sends `{}` as the PATCH body. Harmless no-op server-side, but a zod `.refine` requiring at least one field would make the contract explicit. Same pattern exists in `update_api_key` on main, so it's consistent with repo convention.
2. **`update_user_username` body field name unverified** — `{ username: new_username }` is the documented field, but bogus-id probes can't get past 404 to validate it; the field name remains live-unverified until a real user is renamed. The description already flags probe limits, but not this specific field name.
3. **`userView` drops `updated_at`** — `created_at` is surfaced but `updated_at` isn't; trivial observability nicety for lock/unlock verification (the API response's `updated_at` bump is the cheap confirmation the mutation took).

Everything else checked clean: annotations consistent with the risk table across all 11 tools; password never echoed in any response; owner-refusal path correct; smoke-test guard coverage (6 confirm-guards + owner refusal + 7 bogus-id shapes + regression guard + invalid-email create) all asserted with fail-loud `process.exit(1)`; docs (README tools/scope, gap-analysis Users row + bottom line, gotchas entry) accurate against the code.

**Verdict: both review findings resolved; PR is mergeable as-is.** The three nits above are optional polish — none block.

Issue #6, opened 2026-08-18
Implement IAM SSH Keys, Quotas, JWTs, SAML/SCIM/Security Settings (agentic spec)


gkrost:
# Implement IAM SSH Keys, Quotas, JWTs, SAML/SCIM/Security Settings (agentic spec)

**Direction change 2026-08-18:** previously pre-labeled `wontfix` as an out-of-scope boundary. Direction changed: these IAM surfaces are now in scope and this issue is the self-contained implementation spec, written to be executable end-to-end by an autonomous agent. All Scaleway-behavior claims below that are not already live-verified must be **probed before being relied on** (recipe in each row).

## Objective

Extend the IAM toolset (currently Applications, API keys, Policies/Rules, Permission sets in `src/tools/`) with four independent workstreams, each severable into its own PR if convenient: SSH Keys (CRUD), Quotas (read-only), JWTs (list/get/delete), and org-wide SAML/SCIM/Security Settings.

## API surface (verified against https://www.scaleway.com/en/developers/api/iam/ , 2026-08-18)

All endpoints are `https://api.scaleway.com/iam/v1alpha1/...` — same base URL and bearer-token auth as the existing `iamRequest` path in `src/iamClient.ts`. No new auth plumbing needed.

| Group | Endpoints |
|---|---|
| SSH Keys | `GET /ssh-keys`, `POST /ssh-keys`, `GET /ssh-keys/{id}`, `PATCH /ssh-keys/{id}`, `DELETE /ssh-keys/{id}` |
| Quotas | `GET /quotas` (all, org), `GET /quotas/{name}` |
| JWTs | `GET /jwt`, `GET /jwt/{id}`, `DELETE /jwt/{id}` |
| SAML | `GET /saml`, `POST /saml` (enable), `PATCH /saml` (update), `DELETE /saml` (disable); certificates: `GET /saml/certificates`, `POST /saml/certificates`, `GET /saml/certificates/{id}`, `DELETE /saml/certificates/{id}` |
| SCIM | `GET /scim`, `POST /scim` (enable), `DELETE /scim` (disable); tokens: `GET /scim/tokens`, `POST /scim/tokens`, `DELETE /scim/tokens/{id}` |
| Security Settings | `GET /security-settings`, `PATCH /security-settings` |

## Repo context (read before coding)

| Thing | Where |
|---|---|
[...]

PR #20, opened 2026-08-18
feat(iam): Group management, completing the group_id principal loop (issue #5)

### LINKED ISSUE #5: Implement IAM Groups management: full CRUD + membership (agentic spec)
# Implement IAM Groups management (agentic spec)

**Scope change 2026-08-18:** previously closed as out of scope ("`create_policy`/`set_policy_rules` accept a `group_id` principal, but nothing here creates or manages the group itself"). Direction changed: Groups are now in scope. This is the last IAM gap, and the *easiest* to verify fully live — unlike Users (invitation emails) or SAML/SCIM (org-wide blast radius), Groups are fully reversible, send nothing to anyone, and can be created/deleted freely against a throwaway group. The whole lifecycle below is safe to happy-path test.

## Objective

Make Groups first-class: full CRUD + membership management (add/remove/single/bulk/overwrite), completing the loop so an agent can create a group, add principals, and attach a policy to `group_id` — all within this server. Policy attachment via `group_id` already works (see `policies.ts`); this issue adds the missing group half.

## API surface (verified against https://www.scaleway.com/en/developers/api/iam/groups/ , 2026-08-18)

All `https://api.scaleway.com/iam/v1alpha1/...`, same `iamRequest` path in `src/iamClient.ts`.

| Endpoint | Method | Notes |
|---|---|---|
| `/groups` | GET | list; filters: `order_by` (created_at/updated_at/name × asc/desc), `page`, `page_size`, `organization_id`, `name`, `application_ids[]`, `user_ids[]`, `group_ids[]`, `tag` |
| `/groups` | POST | create; `organization_id` + `name` required (name max 64, **unique per org**), `description` (max 200), `tags` (max 10) |
| `/groups/{group_id}` | GET | full object incl. `user_ids[]`, `application_ids[]` |
| `/groups/{group_id}` | PATCH | update `name`/`description`/`tags` |
| `/groups/{group_id}` | DELETE | 204; **irreversible**, removes group-granted permissions from all members |
| `/groups/{group_id}/add-member` | POST | add ONE user (`user_id`) and/or ONE application (`application_id`) — at most one of each per call |
| `/groups/{group_id}/add-members` | POST | add multiple (`user_ids[]`, `application_ids[]`); existing members are preserved |
| `/groups/{group_id}/members` | PUT | **overwrite** — full-replace of membership (`user_ids[]` + `application_ids[]`, both required); anything omitted is removed |
| `/groups/{group_id}/remove-member` | POST | remove one user and/or one application |

**Group object:** `id, created_at, updated_at, organization_id, name, description, user_ids[], application_ids[], tags[], editable, deletable, managed, all_users, all_applications`.

**Special groups (critical):** Scaleway provisions non-deletable system groups — `managed` groups including the "All Users" (`all_users: true`) and "All Applications" (`all_applications: true`) groups. These are `editable: false` / `deletable: false`. Tools that mutate membership or delete MUST refuse these tool-side (belt-and-braces, matching the owner-refusal pattern in `users.ts`) and/or surface the API's own rejection — but do not rely on the API alone; check `managed`/`all_users`/`all_applications` from a `GET` first.

| Thing | Where |
|---|---|
| IAM client (`iamRequest`, `iamListAll` page/total_count pagination, `withIamErrorHandling`) | `src/iamClient.ts` |
| CRUD + confirm-guard + **owner-refusal** template to copy | `src/tools/users.ts` (`delete_user`'s fetch-then-refuse pattern is exactly what the managed-group guard should look like) |
| `group_id` principal already accepted by policy tools (integration point) | `src/tools/policies.ts` |
| Registration point | `src/index.ts` |
| Smoke harness to extend | `scripts/smoke-test.mjs` |
| Docs to update at the end | `README.md`, `docs/capability-gap-analysis.md` (Groups row), `docs/gotchas.md` |
| Permission sets — **already granted**, no action needed | credential's policy already carries `IAMGroupManager` (org scope, verified 2026-08-18); `IAMGroupReadOnly` exists for read-only identities. Verify with a plain `GET /groups` (expect 200) before assuming. |

Build gate: `npm run build`; verification `node scripts/smoke-test.mjs` (live `.env`).

## Work items — file `src/tools/groups.ts`

Risk classes (repo standard): **[R0]** read · **[R1]** reversible write · **[R2]** access-removing write, `confirm: true` · **[R3]** irreversible, `confirm: true` + explicit warning.

| # | Tool | Op | Risk | Notes |
|---|---|---|---|---|
| 1 | `scaleway_iam_list_groups` | list | R0 | expose filters: `name`, `user_ids[]`, `application_ids[]`, `group_ids[]`, `tag`, `order_by`. Return `total_count` + compact view. |
| 2 | `scaleway_iam_get_group` | get | R0 | full object incl. `user_ids`/`application_ids`/`managed`/`all_users`/`all_applications`. |
| 3 | `scaleway_iam_create_group` | create | R1 | no confirm (reversible, no side effects to anyone). Describe name-uniqueness (and the known create-after-delete uniqueness lag — see gotchas). |
| 4 | `scaleway_iam_update_group` | patch | R1 | name/description/tags. Refuse `managed` groups tool-side. |
| 5 | `scaleway_iam_delete_group` | delete | R3 | confirm. Fetch first; refuse `managed`/`all_users`/`all_applications` tool-side; warn that member permissions are removed irreversibly. |
| 6 | `scaleway_iam_add_group_member` | add-member | R1 | `user_id` XOR-ish `application_id` (validate at least one, both optional). Description: grants the group's attached policies to the member. |
| 7 | `scaleway_iam_add_group_members` | add-members | R1 | `user_ids[]` + `application_ids[]`. |
| 8 | `scaleway_iam_set_group_members` | overwrite (PUT) | R3 | confirm. **Full-replace**: description must say omitted members are removed. Refuse `managed` groups. |
| 9 | `scaleway_iam_remove_group_member` | remove-member | R2 | confirm. Warn that group-granted permissions are revoked from that principal. |

Annotations: R0 → `readOnlyHint: true, idempotentHint: true`; R1 → `readOnlyHint: false, destructiveHint: false, idempotentHint: false` (add is not idempotent); R2/R3 → `destructiveHint: true`.

## Verify-first hypotheses (all resolvable by safe live probes against a throwaway group — no risk)

- **H1 special groups:** does `GET /groups` actually surface the managed "All Users"/"All Applications" groups (confirm `managed`/`all_users`/`all_applications` are real fields on this org), and what does the API return if you DELETE/PUT on one (expect 4xx — record the exact error for the tool description).
- **H2 add-member one-of-each:** confirm `add-member` accepts at most one `user_id` + one `application_id` per call, and the error shape when you pass neither (validation 400) or both is fine (it's "one of each", so both is legal).
- **H3 overwrite semantics:** confirm `PUT /members` full-replaces (create group with 2 members → overwrite with 1 → verify the other is gone from `get`).
- **H4 uniqueness lag:** confirm the name-uniqueness race seen elsewhere (create → delete → immediate re-create same name may 409) applies to groups too; if so, use timestamped names in the smoke test (already the repo convention).
- **H5 list filters:** confirm `name`, `tag`, `user_ids`, `application_ids` filters actually filter server-side (one probe each) — record which need repeated query params vs arrays.

## Verification protocol (FULL happy-path live — this issue has no blast radius)

Extend `scripts/smoke-test.mjs`, all against a throwaway group + a throwaway application, `try/finally` zero-residue cleanup (delete members → delete group → delete app, even on mid-script failure):

1. `list_groups` baseline (expect to see managed groups — record their names/ids).
2. `create_group` (timestamped name) → `get_group` echoes name + empty membership.
3. `list_groups` with `name` filter contains it; with `group_ids` filter returns exactly it.
4. `update_group` rename + description → `get` reflects; `update_group` on a managed group → tool-side refusal (guard test).
5. Create throwaway application → `add_group_member` (application_id) → `get` shows it in `application_ids`.
6. `add_group_members` with an additional id → both present.
7. `set_group_members` overwrite to just one → `get` shows the other removed (proves full-replace; H3).
8. `remove_group_member` → `get` shows empty.
9. `delete_group` → gone from `list_groups`.
10. Guard negatives: `delete_group` without confirm rejected; `remove_group_member` without confirm rejected; `set_group_members` without confirm rejected; `delete_group` on a managed group refused tool-side; `add_group_member` with neither id rejected (schema).
11. Cross-check the integration: `create_policy` with the group's `group_id` as principal succeeds (then delete the policy) — proves the loop closes.
12. Zero residue verified: throwaway group and app gone from their lists.

## Acceptance criteria

1. All 9 tools implemented per the risk table (or a written PR justification for deviation, backed by probe evidence); managed-group refusal present on #4/#5/#8.
2. H1–H5 resolved with recorded evidence in the PR.
3. `npm run build` green; smoke test extended per the protocol and passing end-to-end, zero residue.
4. README tool list + Scope updated (Groups no longer "pending #5"); gap-analysis Groups row flipped to fixed; gotchas.md gains any new findings (special groups, overwrite semantics, uniqueness lag).
5. PR references this issue (`Closes #5`), includes the live evidence matrix (all happy-path, no negative-only caveats — this one can be fully exercised).

## Out of scope

- Nested/sub-groups (Scaleway has no such concept — groups are flat).
- Anything already in other issues (Users #4 done, SSH keys/JWTs/SAML/SCIM #6 done, object/bucket work done).

gkrost:
Closes #5.

## What
9 tools in `src/tools/groups.ts`: `list/get/create/update/delete_group`, `add_group_member(s)`, `set_group_members` (full-replace), `remove_group_member`. Policy tools already accepted `group_id` as a principal - this was the missing group half.

## API surface verification
Live-probed the issue's own table before implementing (it cited the endpoint-specific docs page, unlike issue #6's more general source) - almost entirely correct:

| Finding | Detail |
|---|---|
| Field shapes | `add-member`/`add-members`/`members` (PUT)/`remove-member` all match the issue's spec exactly (flat `user_id`/`application_id`, plural `user_ids[]`/`application_ids[]` as JSON fields) |
| `list_groups` array filters | Issue wrote `group_ids[]=` (bracket-suffixed) - that's **400**. Real format is plain repeated params: `?group_ids=a&group_ids=b`. Same for `user_ids`/`application_ids`. |
| `name` filter | **Exact match**, not substring - `?name=Admin` against "Administrators" returns 0 results. |

## Verification - full happy-path, zero blast radius (per the issue's own protocol)
This is the one IAM issue explicitly designed to be fully live-testable (no invitation emails, no org-wide SSO risk). `scripts/smoke-test.mjs` now runs the complete lifecycle: create → get → list (name + group_ids filters) → rename → add one member (application) → add multiple (additive, proves existing membership preserved) → **overwrite membership** (proves full-replace - the earlier member is gone, H3 resolved) → remove member → **cross-check**: use the group's `group_id` as a real Policy principal via `create_policy`, then clean up → delete, zero residue verified.

Managed-group refusal guard (matching `users.ts`'s owner-refusal pattern) is implemented for `update_group`/`delete_group`/`set_group_members` but **untested live** - this org's only groups are the 3 unmanaged presets (Administrators/Editors/Billing Administrators), none flagged `managed`/`all_users`/`all_applications`.

## Incident (small, self-caused, documented)
While building the smoke test, a scripting bug (`create_policy` called without its required `rules` field) failed mid-`try`-block, and the shared `expectJson` helper's `process.exit(1)` skipped the `finally` cleanup - leaving a throwaway group + Application on the live account until a manual follow-up delete. Same bug class as one hit building issue #6's SSH-key section. Fixed by throwing instead of exiting inside the try block; documented in `docs/gotchas.md` as a standing rule for future cleanup blocks in this file.

## Docs
README tool list + Scope (Groups no longer "pending #5"), `docs/capability-gap-analysis.md` Groups row + bottom line (IAM now has no remaining deliberate scope boundaries), `docs/gotchas.md`.

## Test plan
- [x] `npm run build` clean
- [x] `node scripts/smoke-test.mjs` - full pass, 0 failures, exit 0
- [x] Zero residue verified (group + application both confirmed gone)


src/tools/groups.ts
```diff
@@ -0,0 +1,289 @@
...
+  };
+}
+
+function refuseIfSpecial(g: Group, action: string): string | null {
+  if (g.managed || g.all_users || g.all_applications) {
+    const which = g.all_users ? "the 'All Users' group" : g.all_applications ? "the 'All Applications' group" : "a Scaleway-managed group";
+    return `Refusing to ${action} ${g.id}: it is ${which} (managed=${g.managed}, all_users=${g.all_users}, all_applications=${g.all_applications}). These are provisioned by Scaleway and not meant to be edited/deleted here - the API itself is also expected to reject it.`;
+  }
+  return null;
+}
```
gkrost, line 80: **Special-group guard is narrower than the spec, and never fires on this org's most dangerous groups.**

Issue #5: special groups are `managed` / `all_users` / `all_applications`, and they are also `editable: false` / `deletable: false`. Tools that **mutate membership or delete** must refuse tool-side.

This helper only checks the three boolean flags. The PR itself records that Administrators / Editors / Billing Administrators all have `managed: false` — so this guard does **not** protect those presets. The API already exposes `editable`/`deletable` on the same object (`groupView` even returns them). Use those as well:
- update / set / add / remove → refuse when `editable === false`
- delete → refuse when `deletable === false`

in addition to the three flags. That is what would actually refuse a non-managed preset if Scaleway marks it non-editable.

src/tools/groups.ts
```diff
@@ -0,0 +1,289 @@
...
+      inputSchema: {
+        group_id: groupIdField,
+        name: z.string().min(1).max(64).optional(),
+        description: z.string().max(200).optional(),
+        tags: z.array(z.string()).max(10).optional(),
+      },
+      annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: true, openWorldHint: true },
+    },
+    async ({ group_id, name, description, tags }) => {
+      const current = await iamRequest<Group>(config, "GET", `/groups/${group_id}`);
```
gkrost: **Preflight `GET` is outside `withIamErrorHandling`.** Same on `delete_group` (L178) and `set_group_members` (L258).

`iamRequest` throws `IamApiError` on 404/403. `users.ts` `delete_user` wraps the fetch-then-refuse **inside** `withIamErrorHandling` so a missing id becomes a tool error. Here a 404 on the preflight GET escapes as an unhandled exception (the wrapper only covers the PATCH/DELETE).

Wrap the whole handler body, matching `users.ts`.

src/tools/groups.ts
```diff
@@ -0,0 +1,289 @@
...
+      annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: false, openWorldHint: true },
+    },
+    async ({ group_id, user_id, application_id }) => {
+      const err = requireAtLeastOneMember(user_id, application_id);
+      if (err) return toolError(err);
+      return withIamErrorHandling(async () => {
+        const g = await iamRequest<Group>(config, "POST", `/groups/${group_id}/add-member`, { user_id, application_id });
+        return toolJsonResult(groupView(g), config.MAX_OUTPUT_CHARS);
+      });
+    },
```
gkrost: **Spec miss: membership mutation does not refuse special groups.**

Issue #5: *"Tools that mutate membership or delete MUST refuse these tool-side"* — that is add-member, add-members, set-members, remove-member, and delete. Only `update` / `delete` / `set` call `refuseIfSpecial`. `add_group_member`, `add_group_members`, and `remove_group_member` go straight to the API.

Add the same GET-then-refuse (inside `withIamErrorHandling`) before the POST. Adding a principal to `all_users` / `all_applications` / a `managed` group is exactly the case the belt-and-braces guard exists for.

(Not asking to promote add to `confirm: true` — the issue classifies add as R1.)

gkrost (commented):
Review of #20 (Groups / issue #5). Cannot REQUEST_CHANGES on my own PR, so this is a COMMENT review. Happy-path surface and the list-filter corrections look right; smoke protocol matches the issue.

Findings below: special-group guard is a spec miss and never fires on this org's presets; preflight GET is outside `withIamErrorHandling`; add/remove do not refuse special groups; leftover #4 in the gap-analysis remaining list.

Fixes will land on this branch.

src/tools/groups.ts
```diff
@@ -0,0 +1,289 @@
...
+      description: "Add multiple Users and/or Applications to a Group at once. Existing members are preserved - this is additive, not a replace.",
+      inputSchema: {
+        group_id: groupIdField,
+        user_ids: z.array(z.string().uuid()).optional(),
+        application_ids: z.array(z.string().uuid()).optional(),
+      },
+      annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: false, openWorldHint: true },
+    },
+    async ({ group_id, user_ids, application_ids }) => {
+      if (!user_ids?.length && !application_ids?.length) return toolError("Provide at least one user_id or application_id.");
```
gkrost: Error string says `user_id` / `application_id` (singular) but the fields are `user_ids` / `application_ids`. Same check on `add_group_member` is correctly singular. Copy-paste leftover — will confuse an agent that retries with the wrong names.

docs/capability-gap-analysis.md
```diff
@@ -92,14 +92,15 @@ SSH Keys, JWTs, SAML, SCIM, and Security Settings (formerly listed here as out o
...
-GitHub issue (labeled `scaleway`) rather than left implicit:
+was implemented 2026-08-18 (see the IAM table above), as were Audit Trail's alerting/export machinery,
+bucket configuration, and Groups ([#5](https://github.com/logic-arts-official/scaleway-ops-mcp-server/issues/5),
+completing the `group_id` principal loop policy tools already accepted) - see those tables for what
+works and what's blocked on Scaleway's own API. IAM has no remaining deliberate scope boundaries as
+of this writing - every item originally listed as out of scope has since been implemented. Each
+remaining gap elsewhere is tracked as its own GitHub issue (labeled `scaleway`) rather than left
+implicit:
 [#3](https://github.com/logic-arts-official/scaleway-ops-mcp-server/issues/3) policy clone,
 [#4](https://github.com/logic-arts-official/scaleway-ops-mcp-server/issues/4) Users,
```
gkrost: Leftover. The paragraph above says Users and Groups are implemented and IAM has no remaining deliberate scope boundary, then the "remaining gap" list still links [#4](https://github.com/logic-arts-official/scaleway-ops-mcp-server/issues/4) Users (and dropped #5 but kept #4). Drop #4 from this list the same way #5 was dropped.

gkrost:
Follow-up for the review above: [`5485422`](https://github.com/logic-arts-official/scaleway-ops-mcp-server/commit/5485422).

- Preflight `GET`s on update/delete/set/add/remove now sit inside `withIamErrorHandling` (404/403 become tool errors, matching `users.ts`).
- `refuseIfSpecial` also checks `editable=false` / `deletable=false` — the flags that would actually refuse this org's presets (`managed: false` on Administrators/Editors/Billing).
- Guard now runs on **every** membership mutation (`add_group_member(s)`, `remove_group_member`), not only update/delete/set.
- `add_group_members` error string uses the real field names.
- `update_group` with no fields returns a tool error instead of `PATCH {}`.
- Dropped leftover `#4` from the gap-analysis remaining-issue list.

Did **not** promote add to `confirm: true` — issue #5 classifies it as R1.

`npx tsc --noEmit` exit=0. Smoke test not re-run from this environment (needs live `.env`).

Issue #31, opened 2026-08-19
put_object_tags and put_bucket_tagging disagree on destructiveHint for the identical full-replace pattern


gkrost:
Found in a full-codebase audit (2026-08-19).

**Files:** \`src/tools/objects.ts\` (\`scaleway_s3_put_object_tags\`: \`destructiveHint: true\`) vs. \`src/tools/bucketConfig.ts\` (\`scaleway_s3_put_bucket_tagging\`: \`destructiveHint: false\`).

Both tools do the same thing (full-replace a tag set, no \`confirm\` field, both documented as "REPLACES... does not merge"), yet one is annotated destructive and the other isn't. Since \`destructiveHint\` is one of the signals MCP clients use to decide on confirmation/approval behavior (see the related presigned-URL issue filed alongside this one), this is a real metadata inconsistency between two copy-pasted sibling tools - the same underlying risk (silent data loss on a tag set) is presented differently to whatever is deciding how much scrutiny a call needs, depending purely on which file the tool happens to live in.

**Suggested fix:** pick one and make both consistent - given the full-replace/no-confirm/data-loss-risk profile, \`destructiveHint: true\` on both seems like the more accurate signal.

Issue #26, opened 2026-08-19
generate_presigned_url's put operation mints a write capability but is annotated readOnlyHint: true, no confirm gate


gkrost:
Found in a full-codebase audit (2026-08-19).

**File:** \`src/tools/objects.ts\` - \`presignedUrlSchema\` (no \`confirm\` field), the tool's static \`readOnlyHint: true\` annotation, and its handler.

\`scaleway_s3_generate_presigned_url\` takes \`operation: z.enum(["get", "put"])\` but declares one static annotation, \`readOnlyHint: true\`, covering both branches. For \`operation: "put"\`, the call hands back a URL granting direct, unauthenticated write access to a specific object for up to 7 days (\`expires_in_seconds\` max 604800) - functionally equivalent to exporting a scoped write credential outside the MCP boundary. There's no \`confirm\` field at all, unlike every other "grants a capability" tool in this codebase (\`set_bucket_visibility\`'s public-read, \`put_bucket_website\`, SAML/SCIM enable, etc. all require \`confirm: true\`).

**Failure scenario:** MCP clients (this repo's own \`docs/gotchas.md\` references Claude Code's auto-mode classifier using annotations like this to decide whether a call needs human confirmation) can use \`readOnlyHint\` to skip confirmation for calls declared read-only. Because this tool is declared read-only unconditionally, an agent can mint a 7-day upload URL for any bucket/key and hand it to an external party entirely without human-in-the-loop approval - the one MCP-level safety signal this codebase otherwise relies on consistently doesn't apply to this specific tool's most dangerous branch.

**Suggested fix:** split the annotation by operation (not possible with MCP's static per-tool annotations today, so the practical fix is either always reporting \`readOnlyHint: false\`/\`destructiveHint: true\` for this tool, since it CAN mint a write capability, or requiring \`confirm: true\` specifically when \`operation === "put"\`).

PR #33, opened 2026-08-19
fix(iam): confirm-gate set_policy_rules and refuse self-lockout (#25)

### LINKED ISSUE #25: set_policy_rules can permanently lock out this server's own IAM-management permission, no confirm gate

**File:** \`src/tools/policies.ts\` - \`setRulesSchema\` (no \`confirm\` field) and its handler.

\`scaleway_iam_set_policy_rules\` is a full-replace \`PUT\` with no \`confirm\` field. Its description markets it as "the safe way to change a live policy's grants," referencing the delete+recreate lockout risk already documented in \`docs/gotchas.md\`. But that doc only covers one lockout mode (the policy object briefly not existing during delete+recreate). It doesn't cover this tool's own distinct risk: if the caller reconstructs the \`rules\` array from memory or a stale \`list_policy_rules\` read and omits the rule granting this server's own Application \`IAMPolicyManager\`/\`IAMApplicationManager\`, the call succeeds immediately - no warning, no confirm - and the credential is now permanently unable to call \`set_policy_rules\` (or any IAM policy/application tool) again to fix its own mistake.

**Failure scenario:** Agent calls \`list_policy_rules\`, then calls \`set_policy_rules\` to add a new permission set but reconstructs the array from an incomplete note, dropping the existing \`IAMPolicyManager\`/\`IAMApplicationManager\` rule. The call succeeds. Every subsequent IAM policy/application tool call now fails \`permissions_denied\`, including \`set_policy_rules\` itself - no path back through this server.

**Suggested fix:** require \`confirm: true\` on this tool (matching every other full-replace-with-real-blast-radius tool in the codebase), and/or have the handler warn or refuse when the new rule set doesn't include a rule granting IAMPolicyManager/IAMApplicationManager scoped to this server's own organization/application - similar in spirit to \`scaleway_s3_put_bucket_policy\`'s recommendation to include a "maintain access" statement for the analogous S3-side risk.

gkrost:
Closes #25.

**Stack:** 1/7. Merge this first; #26–#31 are stacked on top.

## Summary
`scaleway_iam_set_policy_rules` is a full-replace `PUT` that was marketed as the safe way to change a live policy, but it had no `confirm` gate. Reconstructing `rules` from a stale `list_policy_rules` read and dropping the `IAMPolicyManager`/`IAMApplicationManager` grant succeeds immediately and permanently locks this credential out of IAM — including this tool itself.

- Require `confirm: true` (same gate as the other full-replace / high-blast-radius tools).
- Before the PUT, GET the current rules. If they include `IAMPolicyManager` or `IAMApplicationManager` and the replacement drops either, refuse with `toolError` and do not write.
- Document this second lockout mode in `docs/gotchas.md` (distinct from the delete+recreate window).

## Test plan
- [ ] Live `confirm` rejection / lockout-refusal not run here (no Scaleway credentials in this environment)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project are documented here.
 
+## Unreleased
+
+- `scaleway_iam_set_policy_rules` now requires `confirm=true` and refuses a full-replace that would drop `IAMPolicyManager`/`IAMApplicationManager` from a policy that currently grants them, so a stale or incomplete rules array cannot permanently lock this credential out of IAM (issue #25).
+
 ## 0.1.1
 
 Security fix: `access_key` (`scaleway_iam_update_api_key`/`scaleway_iam_delete_api_key`) and `jti`
--- a/src/tools/policies.ts
+++ b/src/tools/policies.ts
@@ -2,7 +2,7 @@ import { z } from "zod";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import type { Config } from "../config.js";
 import { iamListAll, iamRequest, withIamErrorHandling } from "../iamClient.js";
-import { toolJsonResult } from "../output.js";
+import { toolJsonResult, toolError } from "../output.js";
 
 interface Rule {
   id: string;
@@ -111,6 +111,12 @@ const setRulesSchema = {
         "append. Call scaleway_iam_list_policy_rules first, add/remove/edit within that array, then pass the " +
         "whole thing back here. This is the safe way to change a live policy's grants: unlike delete+recreate, " +
         "there's no window where the policy (or the permissions it grants its own holder) doesn't exist.",
+    ),
+  confirm: z
+    .literal(true)
+    .describe(
+      "Must be explicitly true. This is a full-replace of live grants: omitting the rule that grants this " +
+        "server IAMPolicyManager/IAMApplicationManager permanently locks the credential out of IAM.",
     ),
 };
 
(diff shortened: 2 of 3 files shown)


src/tools/policies.ts
```diff
@@ -264,15 +270,31 @@ export function registerPolicies(server: McpServer, config: Config) {
...
+        "rule (IAMPolicyManager/IAMApplicationManager) cannot be undone through this server - the credential " +
+        "then cannot call this tool, or any other IAM policy/application tool, to fix itself. Call " +
         "scaleway_iam_list_policy_rules first to get the current array before changing it.",
       inputSchema: setRulesSchema,
       annotations: { readOnlyHint: false, destructiveHint: true, idempotentHint: true, openWorldHint: true },
     },
-    async ({ policy_id, rules }) =>
+    async ({ policy_id, rules, confirm }) =>
       withIamErrorHandling(async () => {
+        const current = await iamRequest<{ rules: Rule[]; total_count: number }>(config, "GET", `/rules?policy_id=${policy_id}`);
```
gkrost, line 285: **[bug]** This lockout guard GETs `/rules?policy_id=…` once with no `page`/`page_size` and never reads `total_count`. Scaleway documents that endpoint as paginated (`page_size` 1–100). If `IAMPolicyManager` / `IAMApplicationManager` is not on page 1, `currentSets` misses them, the drop check is a no-op, and the PUT still full-replaces — the exact false negative #25 was meant to close.

GET failures are already fail-closed via `withIamErrorHandling`, and schema `confirm: true` is enforced before the handler. Those parts are fine.

**Suggestion:** reuse `iamListAll(config, \`/rules?policy_id=${policy_id}\`, "rules")` (same helper `list_policies` already uses for paging). Optionally also keep those permission sets org-scoped to `SCW_ORGANIZATION_ID`, not merely present as names.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project are documented here.
 
+## Unreleased
+
+- `scaleway_iam_set_policy_rules` now requires `confirm=true` and refuses a full-replace that would drop `IAMPolicyManager`/`IAMApplicationManager` from a policy that currently grants them, so a stale or incomplete rules array cannot permanently lock this credential out of IAM (issue #25).
```
gkrost, line 7: **[nit]** Extra blank line between `## Unreleased` and this first bullet. Drop the spare newline so the section matches `## 0.1.1` below.

gkrost (commented):

Two findings on this PR: the lockout GET is a single unpaged `/rules` call (bug), and CHANGELOG has a spare blank line under Unreleased (nit). Follow-up fix incoming.

PR #34, opened 2026-08-19
fix(s3): confirm-gate presigned put URLs and drop readOnlyHint (#26)

### LINKED ISSUE #26: generate_presigned_url's put operation mints a write capability but is annotated readOnlyHint: true, no confirm gate

gkrost:
Closes #26.

**Stack:** 2/7. Base is `fix/25-set-policy-rules-confirm`. Merge after #25, then retarget this PR to `main` (or merge the rest of the stack in order).

## Summary
`scaleway_s3_generate_presigned_url` takes `operation: "get" | "put"` but was annotated `readOnlyHint: true` for both branches. A PUT URL is a time-limited unauthenticated write capability (up to 7 days) with no `confirm` field. MCP clients that skip confirmation for read-only tools could mint one without a human in the loop.

- Annotations are now `readOnlyHint: false` / `destructiveHint: true` (static per-tool; this tool *can* mint a write capability).
- `confirm: true` is required when `operation === "put"`, matching `set_bucket_visibility`'s public-read pattern.
- Smoke test: existing GET call unchanged; added a put-without-confirm `expectError` next to the other S3 confirm guards.

## Test plan
- [ ] Live smoke-test not run here (needs Scaleway credentials)

### DIFF
--- a/scripts/smoke-test.mjs
+++ b/scripts/smoke-test.mjs
@@ -383,6 +383,7 @@ try {
   const vis0 = expectJson(await client.callTool({ name: "scaleway_s3_get_bucket_visibility", arguments: { bucket: cfgBucket } }), "get visibility initial");
   if (vis0.visibility !== "private") { console.error("FAILED: new bucket not private"); process.exit(1); }
   await expectError("scaleway_s3_set_bucket_visibility", { bucket: cfgBucket, visibility: "public-read" }, "public without confirm rejected");
+  await expectError("scaleway_s3_generate_presigned_url", { bucket: cfgBucket, key: "probe.txt", operation: "put" }, "presigned put without confirm rejected");
   console.log("visibility public:", text(await client.callTool({ name: "scaleway_s3_set_bucket_visibility", arguments: { bucket: cfgBucket, visibility: "public-read", confirm: true } })).slice(0, 80));
   const vis1 = expectJson(await client.callTool({ name: "scaleway_s3_get_bucket_visibility", arguments: { bucket: cfgBucket } }), "get visibility public");
   if (vis1.visibility !== "public") { console.error("FAILED: visibility not public after set"); process.exit(1); }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project are documented here.
 ## Unreleased
 
 - `scaleway_iam_set_policy_rules` now requires `confirm=true` and refuses a full-replace that would drop `IAMPolicyManager`/`IAMApplicationManager` from a policy that currently grants them, so a stale or incomplete rules array cannot permanently lock this credential out of IAM (issue #25).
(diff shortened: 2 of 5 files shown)


scripts/smoke-test.mjs
```diff
@@ -383,6 +383,7 @@ try {
   const vis0 = expectJson(await client.callTool({ name: "scaleway_s3_get_bucket_visibility", arguments: { bucket: cfgBucket } }), "get visibility initial");
   if (vis0.visibility !== "private") { console.error("FAILED: new bucket not private"); process.exit(1); }
   await expectError("scaleway_s3_set_bucket_visibility", { bucket: cfgBucket, visibility: "public-read" }, "public without confirm rejected");
+  await expectError("scaleway_s3_generate_presigned_url", { bucket: cfgBucket, key: "probe.txt", operation: "put" }, "presigned put without confirm rejected");
```
gkrost, line 386: **[suggestion]** Good negative case for presigned PUT. Agents will now also be rejected without `confirm=true` on `scaleway_iam_set_policy_rules` and `scaleway_audit_create_export_job`, but smoke never calls those tools (audit still only `list_export_jobs`; the existing “missing confirm rejected BEFORE any API call” loop at ~449 was not extended). A schema/handler regression on those gates would not fail CI/smoke.

**Suggestion:** add both to the schema-reject loop (bogus id / unused name, no `confirm`), same shape as the user/SAML guards.

gkrost (commented):

Suggestion: the new confirm gates on `set_policy_rules` and `create_export_job` have no smoke-test sibling to this presigned-PUT check. Follow-up fix incoming.

PR #35, opened 2026-08-19
fix(audit): confirm-gate create_export_job (#27)

### LINKED ISSUE #27: create_export_job has no confirm gate despite an immediate, hard-to-undo real-data write

**File:** \`src/tools/auditTrailExports.ts\` - export job creation schema (no \`confirm\` field), the tool's \`destructiveHint: false\` annotation.

The tool's own description already states that creating an export job triggers an immediate backfill - \`docs/gotchas.md\` documents a real run producing ~6 days of past daily log objects within seconds - and that \`scaleway_audit_delete_export_job\` does not delete those objects afterward (permanent residue, object-level cleanup is out of this server's scope). That residue is also documented to actively block \`scaleway_s3_delete_bucket\` (\`BucketNotEmpty\`).

Every other tool in the codebase with a comparable "writes real, hard-to-undo state" profile (\`create_user\` sends a real invitation email; SAML/SCIM enable is org-wide) is \`confirm\`-gated. This one, which writes real (potentially sensitive - audit events include who/when/source-IP/resource-id) content into a target bucket with no cleanup path through this server, is not.

**Failure scenario:** Agent calls \`create_export_job\` against a bucket it doesn't realize is broadly readable (e.g. mid-configuration, before a bucket policy is tightened) - six days of Organization audit events land there within seconds, before any confirmation step, and can't be retracted or cleaned up through this server.

**Suggested fix:** add \`confirm: true\` to \`create_export_job\`'s schema, matching the pattern already used for every other real-state-changing tool with comparable blast radius.

gkrost:
Closes #27.

**Stack:** 3/7. Base is `fix/26-presigned-url-put-confirm`.

## Summary
`scaleway_audit_create_export_job` immediately backfills ~6 days of real audit events into the destination bucket. `delete_export_job` does not remove those objects (and the residue blocks `delete_bucket`). Sibling tools with comparable blast radius already require `confirm: true`.

- Add `confirm: confirmSchema` to create, with a description covering the immediate backfill and the lack of object cleanup.
- Tool description now says `Requires confirm=true`.
- One sentence in `docs/gotchas.md`.
- Left `destructiveHint: false` — this is a create, gated by confirm, not a delete.

## Test plan
- [ ] Live export-job create not run here (would write real audit objects)

### DIFF
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -210,7 +210,8 @@ creating a job against an empty throwaway bucket produced six daily log objects
 does not delete already-exported objects (by design, matching every other delete tool in this
 server), so a bucket created purely to test an export job will NOT be empty afterward -
 `scaleway_s3_delete_bucket` fails `BucketNotEmpty` until those objects are removed, which needs a
-raw S3 call since object-level operations are out of this server's scope.
+raw S3 call since object-level operations are out of this server's scope. The tool now requires
+`confirm=true`.
 
 ## Preconfigured alert rules start all-disabled, and the two write tools differ in blast radius
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@ All notable changes to this project are documented here.
 
 ## Unreleased
 
+
 - `scaleway_iam_set_policy_rules` now requires `confirm=true` and refuses a full-replace that would drop `IAMPolicyManager`/`IAMApplicationManager` from a policy that currently grants them, so a stale or incomplete rules array cannot permanently lock this credential out of IAM (issue #25).
 - `scaleway_s3_generate_presigned_url`: `operation: "put"` now requires `confirm=true` and the tool is no longer annotated `readOnlyHint` - a PUT URL exports a time-limited unauthenticated write capability outside the MCP boundary (issue #26).
+- `scaleway_audit_create_export_job` now requires `confirm=true` because creation immediately backfills ~6 days of real audit events into the destination bucket, and `scaleway_audit_delete_export_job` does not remove those objects (#27).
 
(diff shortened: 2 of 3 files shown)


gkrost:
The new `confirm=true` requirement correctly protects creation, but the tool still declares `destructiveHint: false`. Creating an export job immediately writes a backfill of real audit objects to the destination bucket, and the PR documentation explicitly says deletion will not remove them. Please set `destructiveHint: true` so MCP clients expose the same safety signal as the schema and description.

docs/gotchas.md
```diff
@@ -210,7 +210,8 @@ creating a job against an empty throwaway bucket produced six daily log objects
 does not delete already-exported objects (by design, matching every other delete tool in this
 server), so a bucket created purely to test an export job will NOT be empty afterward -
 `scaleway_s3_delete_bucket` fails `BucketNotEmpty` until those objects are removed, which needs a
-raw S3 call since object-level operations are out of this server's scope.
+raw S3 call since object-level operations are out of this server's scope. The tool now requires
```
gkrost, line 213: **[suggestion]** This sentence still says export-job residue “needs a raw S3 call since object-level operations are out of this server’s scope.” Object CRUD (`list_objects` / `delete_objects`) is in this server; that leftover will send operators outside MCP to empty a test bucket.

**Suggestion:** point at `scaleway_s3_list_objects` / `scaleway_s3_delete_objects` (still confirm-gated) instead of “raw S3” / “out of scope.” Keep the `confirm=true` note.

gkrost (commented):

The new `confirm=true` sentence is right; it landed on a leftover claim that object CRUD is out of scope. Follow-up fix incoming.

PR #36, opened 2026-08-19
fix(s3): validate region as fr-par/nl-ams/pl-waw (#28)

### LINKED ISSUE #28: S3 tools' region field is an unvalidated free string, unlike Audit Trail's region enum in the same codebase

**Files:** \`src/tools/buckets.ts\`, \`bucketPolicies.ts\`, \`bucketConfig.ts\`, \`objects.ts\` - all define \`region: z.string().optional()\`. Compare to \`src/tools/auditTrailShared.ts\`, which defines \`auditRegionSchema = z.enum(["fr-par", "nl-ams", "pl-waw"])\` and is reused consistently across every Audit Trail tool.

The S3-side tools never reuse or replicate that enum, so \`region\` accepts any string. A malformed value doesn't fail fast with a clean Zod error - it reaches \`S3Client\` construction, where it either throws the AWS SDK's own generic hostname-validation error (for invalid characters) or, for a validly-formatted-but-wrong region like \`"us-east-1"\`, proceeds to a real DNS/TLS attempt against a nonexistent Scaleway host. Neither is an \`S3ServiceException\`, so \`s3Client.ts\`'s \`handleS3\` (a gap already flagged in \`docs/mvp-readiness-review.md\` for only catching that one error class) re-throws it uncaught instead of returning a clean tool error.

**Failure scenario:** Caller passes \`region: "us-east-1"\` (a real AWS region, not a Scaleway one) to any S3 tool - instead of an immediate "must be one of fr-par/nl-ams/pl-waw" validation error, the call hangs on a DNS lookup and then crashes with a raw, uncaught error with no \`toolError()\` formatting, right where an enum (already written elsewhere in this same codebase) would have caught it instantly.

**Suggested fix:** extract a shared \`scwRegionSchema = z.enum(["fr-par", "nl-ams", "pl-waw"])\` and use it for every S3 tool's \`region\` field, same as Audit Trail already does.

gkrost:
Closes #28.

**Stack:** 4/7. Base is `fix/27-create-export-job-confirm`.

## Summary
S3 tools accepted `region` as a free `z.string()`. A value like `us-east-1` reached `S3Client` construction, did a real DNS/TLS attempt, and was not an `S3ServiceException`, so `handleS3` rethrew it uncaught. Audit Trail already had `z.enum(["fr-par", "nl-ams", "pl-waw"])`.

- New `src/scwRegion.ts` exporting `scwRegionSchema`.
- S3 `region` fields in `buckets.ts`, `bucketPolicies.ts`, `bucketConfig.ts`, `objects.ts` now use it.
- `auditRegionSchema` reuses the same enum so there is one source of truth.
- `SCW_DEFAULT_REGION` / `getS3Client` left unchanged (out of issue scope).

## Test plan
- [ ] Invalid-region Zod rejection not live-tested here

### DIFF
--- a/src/tools/objects.ts
+++ b/src/tools/objects.ts
@@ -15,9 +15,10 @@ import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
 import type { Config } from "../config.js";
 import { getS3Client, handleS3 } from "../s3Client.js";
 import { toolJsonResult, toolError } from "../output.js";
+import { scwRegionSchema } from "../scwRegion.js";
 
 const bucketField = z.string().min(3).describe("Bucket name, e.g. 'payments-backups'.");
-const regionField = z.string().optional().describe("Region the bucket lives in. Defaults to the server's configured region (fr-par).");
+const regionField = scwRegionSchema.optional().describe("Region the bucket lives in. Defaults to the server's configured region (fr-par).");
 const keyField = z.string().min(1).describe("Object key (the full path within the bucket), e.g. 'invoices/2026-08/inv-001.pdf'.");
 
 /** URI-encode each path segment of a key but keep the '/' separators literal - required by S3's CopySource format. */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project are documented here.
 - `scaleway_iam_set_policy_rules` now requires `confirm=true` and refuses a full-replace that would drop `IAMPolicyManager`/`IAMApplicationManager` from a policy that currently grants them, so a stale or incomplete rules array cannot permanently lock this credential out of IAM (issue #25).
 - `scaleway_s3_generate_presigned_url`: `operation: "put"` now requires `confirm=true` and the tool is no longer annotated `readOnlyHint` - a PUT URL exports a time-limited unauthenticated write capability outside the MCP boundary (issue #26).
 - `scaleway_audit_create_export_job` now requires `confirm=true` because creation immediately backfills ~6 days of real audit events into the destination bucket, and `scaleway_audit_delete_export_job` does not remove those objects (#27).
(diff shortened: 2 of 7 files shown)


src/tools/objects.ts
```diff
@@ -15,9 +15,10 @@ import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
 import type { Config } from "../config.js";
 import { getS3Client, handleS3 } from "../s3Client.js";
 import { toolJsonResult, toolError } from "../output.js";
+import { scwRegionSchema } from "../scwRegion.js";
 
 const bucketField = z.string().min(3).describe("Bucket name, e.g. 'payments-backups'.");
-const regionField = z.string().optional().describe("Region the bucket lives in. Defaults to the server's configured region (fr-par).");
+const regionField = scwRegionSchema.optional().describe("Region the bucket lives in. Defaults to the server's configured region (fr-par).");
```
gkrost, line 21: **[suggestion]** Explicit `region: "us-east-1"` now fails at Zod, which is the #28 fix. The common path still omits `region` and `getS3Client` falls back to `config.SCW_DEFAULT_REGION`, which remains `z.string().default("fr-par")` in `src/config.ts`. A typo / zone (`fr-par-1`) / AWS region in the env still constructs `https://s3.<bad>.scw.cloud`, and the DNS/TLS error is not an `S3ServiceException`, so `handleS3` rethrows it uncaught — the same failure mode, via the default.

**Suggestion:** type `SCW_DEFAULT_REGION` with `scwRegionSchema` so a bad env fails at `loadConfig()`.

gkrost (commented):

Tool `region` is now an enum; the server default is still a free string, so #28’s failure mode still happens via `SCW_DEFAULT_REGION`. Follow-up fix incoming.

PR #37, opened 2026-08-19
fix(s3): fail fast when get_object exceeds size ceiling (#29)

### LINKED ISSUE #29: get_object buffers/encodes the full object before MAX_OUTPUT_CHARS truncation applies

**Files:** \`src/tools/objects.ts\` - \`get_object\` handler, \`bodyToBuffer\`, \`isUtf8Text\`; truncation site in \`src/output.ts\` (\`JSON.stringify\` runs before \`truncate()\`).

Unlike \`put_object\`, which checks \`body.byteLength > config.MAX_PUT_OBJECT_BYTES\` and fails fast before any upload work, \`get_object\` has no equivalent pre-flight size check. \`bodyToBuffer\` reads the entire object into memory regardless of size, \`isUtf8Text\` does a full decode/re-encode/compare pass over that whole buffer, the content is then base64- or utf8-encoded into the result, and only then does \`toolJsonResult\` run \`JSON.stringify\` over the whole thing before \`truncate()\` cuts it down to \`MAX_OUTPUT_CHARS\` (default 25,000 characters).

**Failure scenario:** Calling \`get_object\` on a multi-hundred-MB or multi-GB object (nothing prevents this - the tool description only suggests presigned URLs "for large objects" as a recommendation, not an enforced limit) buffers the full object in memory, pays a full UTF-8 validation pass and a full base64 encode (~33% size inflation for binary content), and a full JSON.stringify copy - all of which is discarded except the first ~25KB. Worst case this is a memory-pressure/latency problem on the MCP server process for a call that could instead fail in milliseconds with a clear "object too large, use generate_presigned_url" error, mirroring the guard \`put_object\` already has.

**Suggested fix:** add a \`head_object\`-style size pre-check (or use the \`Content-Length\` already available on the GetObject response headers) before buffering the body, and fail fast with a clear error above some reasonable size ceiling - same pattern as \`put_object\`'s existing \`MAX_PUT_OBJECT_BYTES\` guard.

gkrost:
Closes #29.

**Stack:** 5/7. Base is `fix/28-s3-region-enum`.

## Summary
`scaleway_s3_get_object` buffered the entire object, ran a full UTF-8 validation pass, encoded it, then truncated to `MAX_OUTPUT_CHARS`. A large object could OOM / latency-spike the MCP process. `put_object` already fails fast on `MAX_PUT_OBJECT_BYTES`.

- New `MAX_GET_OBJECT_BYTES` (default 5 MB) in `config.ts`, `.env.example`, and the README config table.
- Check `ContentLength` on the GetObject response; HeadObject fallback; fail closed if size is still unknown.
- Size 0 still succeeds. Over-ceiling calls return `toolError` pointing at `scaleway_s3_generate_presigned_url` and never call `bodyToBuffer` / `isUtf8Text`.

## Test plan
- [ ] Oversize `get_object` not live-tested here


gkrost:
The size check occurs after `GetObjectCommand` has opened the object response. For an oversized object, that can still initiate the data transfer and leaves the returned body stream unconsumed, so this is not a true fail-fast guard. Issue `HeadObjectCommand` first, reject based on its `ContentLength`, then call `GetObjectCommand` only within the configured limit.

src/tools/objects.ts
```diff
@@ -170,15 +170,31 @@ export function registerObjects(server: McpServer, config: Config) {
...
+        if (sizeBytes === undefined) {
+          const head = await client.send(new HeadObjectCommand({ Bucket: bucket, Key: key }));
+          sizeBytes = head.ContentLength;
+        }
+        if (sizeBytes === undefined) {
+          return toolError(
+            "Object size could not be determined. Use scaleway_s3_generate_presigned_url or scaleway_s3_head_object.",
+          );
+        }
+        if (sizeBytes > config.MAX_GET_OBJECT_BYTES) {
```
gkrost, line 192: **[bug]** The size cap runs only after `GetObjectCommand` has already started the object body. On this oversize path (and the unknown-size `toolError` just above) the handler returns without consuming or destroying `res.Body`. AWS SDK v3 leaves that HTTP stream open, so a multi-GB object still transfers into an unread socket — the opposite of the “fail fast, never call `bodyToBuffer`” claim.

`ContentLength === 0` empty objects are fine; fail-closed when both Get and Head omit `ContentLength` is also correct. Config wiring for `MAX_GET_OBJECT_BYTES` is complete.

**Suggestion:** `HeadObjectCommand` first, refuse on `head.ContentLength` / unknown size, and only then `GetObjectCommand`. If a Get has already started, `destroy()` the body on any early return.

gkrost (commented):

Bug: the size cap runs after `GetObject` has already started the body, so an oversize object still streams and the unread body is leaked. Follow-up fix incoming.

PR #38, opened 2026-08-19
fix(iam): warn that update_user tags replace the full list (#30)

### LINKED ISSUE #30: update_user's tags field is missing the replace-not-merge warning its sibling tools carry

**Files:** \`src/tools/users.ts\` (\`update_user\` - no warning) vs. \`src/tools/applications.ts\` and \`src/tools/policies.ts\` (both carry an explicit "REPLACES the full tag list... pass the complete set you want, not just the ones you're adding" warning for the identical \`PATCH .../{id}\` + \`tags\` shape).

\`docs/gotchas.md\` confirms this replace-not-merge behavior specifically for \`PATCH .../applications/{id}\` and \`PATCH .../policies/{id}\`, and both of those tools' descriptions quote it. \`update_user\` hits the same IAM API family (\`PATCH /users/{id}\`) with the same \`tags: z.array(z.string()).optional()\` shape, but its description only says "Only passed fields change" - with no mention that passing \`tags\` at all replaces the complete set.

**Failure scenario:** A User has \`tags: ["dept=finance", "cost-center=104"]\`. Caller invokes \`update_user\` intending to *add* \`"onboarding=pending"\` and passes only \`tags: ["onboarding=pending"]\`, reasonably assuming (since the description doesn't say otherwise, unlike its siblings) this is additive - silently drops the other two tags with no warning in the tool description that would have prevented the mistake.

**Suggested fix:** copy the same replace-not-merge warning already present in \`applications.ts\`/\`policies.ts\`'s \`tags\` field description into \`users.ts\`'s \`update_user\` schema.

gkrost:
Closes #30.

**Stack:** 6/7. Base is `fix/29-get-object-size-cap`.

## Summary
`scaleway_iam_update_user`'s `tags` field had no replace-not-merge warning. `update_application` and `update_policy` already warn that passing `tags` replaces the complete set on the same IAM `PATCH .../{id}` shape.

- Copy the sibling warning onto `update_user` tags: locked vocabulary, `=` vs `:`, max 10, and "REPLACES the full tag list — pass the complete set you want".
- Tool-level "Only passed fields change" is unchanged.

## Test plan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project are documented here.
 - `scaleway_audit_create_export_job` now requires `confirm=true` because creation immediately backfills ~6 days of real audit events into the destination bucket, and `scaleway_audit_delete_export_job` does not remove those objects (#27).
 - Validate S3 `region` as `fr-par`/`nl-ams`/`pl-waw` instead of a free-form string, so a bad region fails at schema validation rather than an uncaught DNS/TLS error (#28).
 - `scaleway_s3_get_object` fails fast when the object exceeds `MAX_GET_OBJECT_BYTES` (default 5 MB) instead of buffering it into memory (#29).
+- Warn that `scaleway_iam_update_user` tags replace the full list (#30)
 
 ## 0.1.1
 
--- a/src/tools/users.ts
+++ b/src/tools/users.ts
@@ -115,7 +115,18 @@ export function registerUsers(server: McpServer, config: Config) {
         "Update a User's mutable profile fields (tags; name/email where the API allows - invalid fields are rejected with the API's own validation message). Only passed fields change.",
       inputSchema: {
         user_id: z.string().uuid(),
-        tags: z.array(z.string()).optional(),
+        tags: z
+          .array(z.string())
+          .max(10)
+          .optional()
+          .describe(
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project are documented here.
 - `scaleway_audit_create_export_job` now requires `confirm=true` because creation immediately backfills ~6 days of real audit events into the destination bucket, and `scaleway_audit_delete_export_job` does not remove those objects (#27).
 - Validate S3 `region` as `fr-par`/`nl-ams`/`pl-waw` instead of a free-form string, so a bad region fails at schema validation rather than an uncaught DNS/TLS error (#28).
 - `scaleway_s3_get_object` fails fast when the object exceeds `MAX_GET_OBJECT_BYTES` (default 5 MB) instead of buffering it into memory (#29).
+- Warn that `scaleway_iam_update_user` tags replace the full list (#30)
```
gkrost, line 13: **[nit]** This #30 bullet is a fragment (“Warn that … (#30)”) and has no period, unlike the complete sentences for #25–#29 and #31.

**Suggestion:** match the surrounding style, e.g. “`scaleway_iam_update_user`’s `tags` field now warns that the array replaces the full list, not a merge (#30).”

gkrost (commented):

Nit on the Unreleased changelog bullet wording. Follow-up fix incoming.

Issue #51, opened 2026-08-19
SCW_ORGANIZATION_ID/SCW_PROJECT_ID env vars are the only ID-shaped values with no format constraint


gkrost:
Found in a follow-up audit after issues #25-#31 were fixed (2026-08-19).

**File:** \`src/config.ts:7-8\`:
\`\`\`js
SCW_ORGANIZATION_ID: z.string().min(1, "SCW_ORGANIZATION_ID is required"),
SCW_PROJECT_ID: z.string().min(1, "SCW_PROJECT_ID is required"),
\`\`\`
Every per-call tool-input ID field in the codebase is now \`.uuid()\`-or-regex-constrained, but these two env-var-sourced config values - interpolated into essentially every IAM request path/query, including the org-nested SAML/SCIM/Security-Settings paths - are validated only as non-empty strings. Severity is low: these are operator-set once at startup, not per-call agent input (a different, lower-risk trust boundary than the \`access_key\`/\`jti\` class of bug already fixed), and a typo'd value would most likely just produce a clean 404/403 rather than anything dangerous. Still, tightening to \`.uuid()\` would be a one-line, zero-risk fix consistent with the rest of the codebase's philosophy of failing fast at config/schema time - the same rationale already applied to \`SCW_DEFAULT_REGION\`'s enum.

Same-severity aside noticed alongside this: \`securitySettings.ts\`'s three duration-string fields (\`grace_period_duration\`, \`max_login_session_duration\`, \`max_api_key_expiration_duration\`) accept any string with no format regex - low severity since a bad value just surfaces as a clean Scaleway 400, not a crash, but same theme.

**Suggested fix:** change both env var schemas to \`.uuid()\`, matching every other ID field in the codebase.

Issue #44, opened 2026-08-19
set_policy_rules lockout guard checks permission-set names only, not rule scope


gkrost:
Found in a follow-up audit after issue #25 was fixed (2026-08-19).

**File:** \`src/tools/policies.ts\`, the lockout-detection in \`set_policy_rules\`:
\`\`\`js
const currentSets = new Set(currentRules.flatMap((r) => r.permission_set_names));
const incomingSets = new Set(rules.flatMap((r) => r.permission_set_names));
\`\`\`

This is a plain string-membership check across \`permission_set_names\`, flattened out of every rule - it never checks whether the rule carrying \`"IAMPolicyManager"\`/\`"IAMApplicationManager"\` in the *incoming* array is scoped with \`organization_id\` (the correct org-wide scope these permission sets require) versus \`project_ids\`. \`ruleSchema\` still doesn't enforce "exactly one of \`project_ids\`/\`organization_id\`" (already tracked as open in \`docs/mvp-readiness-review.md\`), so nothing stops a rule like \`{permission_set_names: ["IAMPolicyManager"], project_ids: [...]}\` from reaching this guard undetected.

**Failure scenario:** an agent reconstructing the rules array (e.g. to also add a project-scoped permission set) accidentally moves the IAM-management rule's scope from \`organization_id\` to \`project_ids\` while keeping the name string. \`incomingSets.has("IAMPolicyManager")\` is still \`true\`, so the guard does not refuse, and the PUT is sent - whether this actually preserves working org-wide IAM management depends on unverified Scaleway server-side behavior. The tool's implicit claim of "correctly compares current vs. incoming grants" is only true at the name level, not the scope level that actually determines whether the grant still works.

**Suggested fix:** the lockout check should compare on (permission_set_name, scope_type, scope_id) tuples, not bare names - and/or fix the underlying \`ruleSchema\` gap (exactly-one-of project_ids/organization_id) this finding depends on.

PR #52, opened 2026-08-19
fix(iam): refuse delete_policy when it would lock out IAM (#43)

### LINKED ISSUE #43: delete_policy has no lockout guard, unlike its just-patched sibling set_policy_rules

**File:** \`src/tools/policies.ts\` - \`scaleway_iam_delete_policy\` handler (the tool immediately after \`set_policy_rules\` in the same file).

Issue #25 hardened \`set_policy_rules\` specifically because dropping \`IAMPolicyManager\`/\`IAMApplicationManager\` from the policy that grants this server's own credential IAM management is a permanent, unrecoverable-through-this-server lockout. \`delete_policy\` - the more blunt way to cause identically the same outcome - has no equivalent protection; it only requires \`confirm: true\`.

**Failure scenario:** this server's credential holds IAMPolicyManager/IAMApplicationManager via a single policy \`P\`. An agent (acting on stale context, or asked to "clean up unused policies") calls \`scaleway_iam_delete_policy({policy_id: P, confirm: true})\`. The delete succeeds immediately. Every IAM tool this server exposes now fails \`permissions_denied\` - with no recovery path through this MCP server, exactly the scenario #25's fix exists to prevent, just reached through the sibling tool that never got hardened.

The same root gap conditionally applies to \`scaleway_iam_delete_group\`/\`remove_group_member\`/\`set_group_members\` if this server's own Application is ever made a member of a group that a policy grants IAM management to via \`group_id\` - none of \`refuseIfSpecial\`'s checks (managed/all_users/all_applications/editable/deletable) inspect permission-set content.

**Suggested fix:** before deleting a policy, check whether it currently grants IAMPolicyManager/IAMApplicationManager to this server's own credential (same detection logic \`set_policy_rules\` already has) and refuse (or require an extra explicit acknowledgment) if so - same protection, applied to the other way to reach the same outcome.

gkrost:
Closes #43.

**Stack:** 1/9. Merge this first; #44–#51 are stacked on top.

## Summary
`set_policy_rules` now refuses dropping `IAMPolicyManager`/`IAMApplicationManager`. `delete_policy` was the blunter way to the same lockout and only required `confirm: true`.

- Shared `grantsIamManagement()` helper.
- Before DELETE, page current rules via `iamListAll`. If they grant those permission sets, `toolError` and do not delete — even with confirm.

## Test plan
- [ ] Live delete of a throwaway non-IAM policy still works (smoke already deletes throwaway policies)


src/tools/policies.ts
```diff
@@ -304,12 +317,25 @@ export function registerPolicies(server: McpServer, config: Config) {
...
+        "delete is refused even with confirm, because it would permanently lock out IAM management for whoever " +
+        "this policy currently grants it to (including this server, if this is its own IAM policy). Change the " +
+        "rules first with scaleway_iam_set_policy_rules, or leave the policy in place.",
       inputSchema: deleteSchema,
       annotations: { readOnlyHint: false, destructiveHint: true, idempotentHint: false, openWorldHint: true },
     },
     async ({ policy_id }) =>
       withIamErrorHandling(async () => {
+        const currentRules = await iamListAll<Rule>(config, `/rules?policy_id=${policy_id}`, "rules");
+        if (grantsIamManagement(currentRules)) {
```
gkrost, line 363: `delete_policy`'s guard only covers deleting the *policy*. It doesn't cover deleting the policy's *principal* instead, which reaches the same lockout without ever calling this handler.

`scaleway_iam_delete_application`'s own tool description already says it "detaches every policy scoped to it" - DETACH, not delete - and `scaleway_iam_delete_group`'s description documents the identical consequence for a policy naming that group's `group_id`. So an application/group that holds this server's own IAM-management policy can still be permanently locked out by calling `delete_application`/`delete_group` on the principal, even after this PR - `grantsIamManagement` is never consulted on that path because neither `applications.ts` nor `groups.ts` fetches the attached policy's rules before deleting.

Concretely: if `ops-scaleway-mcp-admin` (or whatever this server's own Application is named, per the naming convention in `applications.ts`) holds the policy that grants it `IAMPolicyManager`, `scaleway_iam_delete_application` on that Application id detaches (orphans) that policy and disables this server's own IAM management, with no guard anywhere - the exact scenario #43 is about, reached from one hop over.

Given the PR title is "refuse delete_policy when it would lock out IAM," I think this is worth closing in the same pass rather than leaving as a known gap - pushing a fix now that extends the same `grantsIamManagement` check to `delete_application`/`delete_group` via a small shared helper.

gkrost (commented):
**Finding (not an inline comment - `scripts/smoke-test.mjs` isn't touched by this diff, but this PR breaks it):**

`scripts/smoke-test.mjs` line 717-723 creates a throwaway Policy with an **org-scoped `IAMPolicyManager` rule** (`organization_id: creds.SCW_ORGANIZATION_ID`) attached to a throwaway group, purely to exercise the `group_id`-as-principal path (`create_policy (group_id principal)`), then calls `delete_policy` with `confirm: true` to clean it up.

After this PR, that `delete_policy` call is refused by the new lockout guard - it's exactly the case the guard exists to catch. The test doesn't assert on the result (`const deletedPolicy = await client.callTool(...); console.log(...)`, no `isError` check), so it won't hard-fail the script, but the policy is never actually deleted. Worse, after it becomes an "existing policy that grants IAMPolicyManager," `set_policy_rules` can never strip that grant either (same guard, org-scoped). Every live run of the smoke test from now on leaks one more permanently-undeletable, org-wide-`IAMPolicyManager`-granting policy into the real account, orphaned from a group that gets deleted a few lines later in `finally`.

Pushing a fix that swaps the permission set for this specific cross-check to something that isn't `IAMPolicyManager`/`IAMApplicationManager` (`AuditTrailReadOnly`, also org-scoped per `docs/gotchas.md`) - the test only needs to prove `group_id` works as a principal, not exercise the lockout guard.

**Second finding, inline on `src/tools/policies.ts`:** `delete_policy`'s guard only covers deleting the policy itself - `scaleway_iam_delete_application`/`scaleway_iam_delete_group` reach the identical lockout by deleting the policy's *principal* instead (both tools' own descriptions already say this detaches, not deletes, the attached policy), and neither goes through `grantsIamManagement`. See the inline comment for detail. Pushing a fix that extends the same check to both.

PR #56, opened 2026-08-19
fix: omit structuredContent when over MAX_OUTPUT_CHARS (#47)

### LINKED ISSUE #47: structuredContent always carries the full untruncated payload - MAX_OUTPUT_CHARS truncation is cosmetic
Found in a follow-up audit after issues #25-#31 were fixed (2026-08-19). This is cross-cutting - affects every tool in the codebase, not just one domain.

**File:** \`src/output.ts:17-23\`
\`\`\`js
export function toolJsonResult(data: unknown, maxChars: number): CallToolResult {
  return {
    content: [{ type: "text", text: truncate(JSON.stringify(data, null, 2), maxChars) }],
    structuredContent: data as Record<string, unknown>,
  };
}
\`\`\`
Only \`content[0].text\` is truncated; \`structuredContent\` gets the complete, untruncated \`data\` object and is a top-level field of \`CallToolResult\` that gets fully serialized into the response regardless of \`MAX_OUTPUT_CHARS\`. No tool declares an \`outputSchema\`, so nothing else bounds it either.

**Failure scenario:** \`scaleway_audit_list_events\` with \`max_pages: 50\` (the schema max) fetches up to 5,000 events. The \`text\` mirror gets correctly truncated to 25,000 chars (default \`MAX_OUTPUT_CHARS\`), but \`structuredContent\` still carries all ~5,000 compacted event objects - silently defeating the documented purpose of \`MAX_OUTPUT_CHARS\` ("Truncation limit for tool responses," README.md) while the visible \`truncated\` flag on the text side implies the response was actually bounded. Same mechanism hits every large \`toolJsonResult\` call in the codebase (big \`list_objects\` pages, \`list_alert_rules\`, \`list_export_jobs\`, etc). Related, same root cause: \`truncate()\` slices pretty-printed JSON mid-structure and appends a human-readable suffix, so a truncated \`text\` field is no longer valid JSON even where truncation does apply.

**Suggested fix:** either truncate/omit \`structuredContent\` above the same \`maxChars\` ceiling (e.g. only include it when the untruncated JSON already fits), or drop \`structuredContent\` entirely and rely on the text mirror, which is the only field this codebase currently treats as size-bounded.

gkrost:
Closes #47.

**Stack:** 5/9. Base is `fix/46-put-bucket-policy-confirm`.

## Summary
`toolJsonResult` truncated only the text mirror. `structuredContent` always carried the full object, so `MAX_OUTPUT_CHARS` was cosmetic.

- Include `structuredContent` only when untruncated pretty JSON fits under the ceiling.
- Over the ceiling, omit it so the wire payload is the truncated text.

## Test plan

### DIFF
--- a/src/output.ts
+++ b/src/output.ts
@@ -14,12 +14,17 @@ export function toolResult(text: string, maxChars: number): CallToolResult {
   };
 }
 
-/** Build a successful CallToolResult carrying structured JSON, mirrored as text. */
+/** Build a successful CallToolResult carrying structured JSON, mirrored as text.
+ * structuredContent is omitted when the untruncated JSON exceeds maxChars. */
 export function toolJsonResult(data: unknown, maxChars: number): CallToolResult {
-  return {
-    content: [{ type: "text", text: truncate(JSON.stringify(data, null, 2), maxChars) }],
-    structuredContent: data as Record<string, unknown>,
+  const text = JSON.stringify(data, null, 2);
+  const result: CallToolResult = {
+    content: [{ type: "text", text: truncate(text, maxChars) }],
   };
+  if (text.length <= maxChars) {
+    result.structuredContent = data as Record<string, unknown>;
+  }
+  return result;
 }
 
 /** Build an error CallToolResult. */
--- a/src/output.ts
+++ b/src/output.ts
@@ -15,16 +15,23 @@ export function toolResult(text: string, maxChars: number): CallToolResult {
 }
 
 /** Build a successful CallToolResult carrying structured JSON, mirrored as text.
- * structuredContent is omitted when the untruncated JSON exceeds maxChars. */
+ * When the untruncated JSON exceeds maxChars, structuredContent carries a truncation
+ * marker instead of the real data - never the field silently missing - so a caller
+ * that reads structuredContent (rather than parsing the text mirror) gets an explicit,
+ * typed signal that it was cut, instead of something indistinguishable from "no
+ * structured content for this tool". */
 export function toolJsonResult(data: unknown, maxChars: number): CallToolResult {
   const text = JSON.stringify(data, null, 2);
-  const result: CallToolResult = {
+  const truncated = text.length > maxChars;
+  return {
     content: [{ type: "text", text: truncate(text, maxChars) }],
+    structuredContent: truncated
+      ? {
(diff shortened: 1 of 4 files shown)


src/output.ts
```diff
@@ -14,12 +14,17 @@ export function toolResult(text: string, maxChars: number): CallToolResult {
...
-    content: [{ type: "text", text: truncate(JSON.stringify(data, null, 2), maxChars) }],
-    structuredContent: data as Record<string, unknown>,
+  const text = JSON.stringify(data, null, 2);
+  const result: CallToolResult = {
+    content: [{ type: "text", text: truncate(text, maxChars) }],
   };
+  if (text.length <= maxChars) {
+    result.structuredContent = data as Record<string, unknown>;
+  }
+  return result;
```
gkrost: **Two real gaps, both fixed in a follow-up commit on this branch:**

1. **Truncation is signaled only in the text mirror, not in `structuredContent`.** When the JSON exceeds `maxChars`, this correctly stops shipping the untruncated object, but it does so by *omitting the field entirely* rather than replacing it with something that says why. A client that reads `structuredContent` preferentially (the MCP-idiomatic path for a typed/programmatic caller) sees only `undefined` — indistinguishable from "this tool never returns structured content" — and has to fall back to parsing prose out of `content[0].text` to discover that (a) truncation happened and (b) how much was cut. The safer shape is to always populate `structuredContent`, with a small explicit marker object (`{ truncated: true, note: ... }`) on the over-the-ceiling path, so a caller keying off `structuredContent` gets an unambiguous, typed signal instead of silence.
2. **No test exercises the new behavior.** The diff doesn't touch `scripts/smoke-test.mjs` (this repo's only test harness) at all, unlike sibling PRs #55/#57 in the same stack which each added an `expectError` case for their own gate. There's currently no live evidence that `structuredContent` is actually absent over the ceiling, or that the non-truncated path is unaffected.

Pushed a commit to `fix/47-structured-content-cap` addressing both: `toolJsonResult` now always returns a `structuredContent` object (a truncation marker when over the ceiling, the real data otherwise), and `scripts/smoke-test.mjs` gets a dedicated second client with a tiny `MAX_OUTPUT_CHARS` override to prove both branches live.

PR #57, opened 2026-08-19
fix(s3): confirm-gate full-replace tagging and CORS puts (#48)

### LINKED ISSUE #48: put_bucket_tagging/put_object_tags/put_bucket_cors inconsistently confirm-gated for the same full-replace risk

**Files:** \`src/tools/bucketConfig.ts\` (\`put_bucket_tagging\`, \`put_bucket_cors\`), \`src/tools/objects.ts\` (\`put_object_tags\`).

All three silently drop any existing tag/rule not included in the call (documented as full-replace in their own descriptions), yet none has a \`confirm\` field. \`put_bucket_tagging\` got \`destructiveHint: true\` in the recent fix for issue #31, but that fix was exactly a one-line annotation flip - no \`confirm\` was added, and its sibling \`delete_bucket_tagging\` (arguably *lower* stakes than an easy-to-get-subtly-wrong partial replace) already requires \`confirm: true\`. \`put_bucket_cors\` has identical full-replace risk and real service impact (breaks a web app's cross-origin requests) and wasn't touched by #31 at all - still \`destructiveHint: false\`, no confirm.

**Failure scenario:** caller wants to add one tag to a bucket that already carries several (\`env\`, \`owner\`, \`cost-center\`) and calls \`put_bucket_tagging\` with only the new tag - nothing forces a prior \`get_bucket_tagging\` read, so the existing tags are gone immediately with no confirmation step.

**Suggested fix:** add \`confirm: z.literal(true)\` to all three, matching the pattern already used for \`delete_bucket_tagging\` and every other full-replace-with-real-consequences tool in the codebase.

gkrost:
Closes #48.

**Stack:** 6/9. Base is `fix/47-structured-content-cap`.

## Summary
`put_bucket_tagging`, `put_object_tags`, and `put_bucket_cors` are full-replace with no confirm. `delete_bucket_tagging` already required it; CORS was still `destructiveHint: false`.

- `confirm: true` on all three.
- `put_bucket_cors` is `destructiveHint: true`.
- Smoke-test: without-confirm guards + live puts pass confirm.

## Test plan


src/tools/objects.ts
```diff
@@ -119,6 +119,7 @@ const putObjectTagsSchema = {
         "object's entire tag set, it does not merge - call scaleway_s3_get_object_tags first if you need to keep " +
         "existing tags, same replace-not-merge semantics as scaleway_s3_put_bucket_policy.",
     ),
+  confirm: z.literal(true).describe("Must be explicitly true. Replacement is immediate and does not merge (omitted tags are dropped)."),
```
gkrost, line 151: **Real gap in the same file, same pattern, not covered by this PR:** `scaleway_s3_put_object` and `scaleway_s3_copy_object` (both further up in this file) can silently overwrite an existing object's content with no `confirm` gate at all, while `scaleway_s3_delete_object`/`scaleway_s3_delete_objects` in this exact file require `confirm=true` for the same class of loss ("immediate and irreversible ... unless the bucket has versioning enabled, which this server does not manage"). Overwriting an existing key is exactly as destructive as deleting it — the old bytes are gone the same way, through the same missing-versioning gap — but today it's ungated and `destructiveHint` is even set to `false` on both tools.

This PR already extends the confirm-gate pattern to `put_object_tags` right here in this file; `put_object`/`copy_object` are the two full-content-replace operations it didn't reach.

Pushed a commit to `fix/48-tagging-cors-confirm` closing this the same way `scaleway_s3_set_bucket_versioning`/`scaleway_s3_set_bucket_visibility` already handle their own conditional risk in the sibling `bucketConfig.ts`: HEAD the destination key first, and only require `confirm=true` on the branch that would actually overwrite something (mirrors `scaleway_s3_get_object`'s existing HEAD-first pattern, so a plain new-object upload/copy needs no extra flag). Also annotated both `destructiveHint: true` to match `delete_object`, and added smoke-test coverage for the new guard on both tools.

PR #59, opened 2026-08-19
fix(iam): cap and document create_user tags like update_user (#50)

### LINKED ISSUE #50: create_user's tags field is missing the max(10) cap and locked-vocabulary description every sibling has

**File:** \`src/tools/users.ts\`, \`create_user\` schema: \`tags: z.array(z.string()).optional()\`.

Every other tags field in this codebase - \`applications.ts\` update, \`policies.ts\` update, \`groups.ts\` create/update, and even \`users.ts\`'s own \`update_user\` two tools down - carries \`.max(10)\` plus the locked-vocabulary \`.describe()\` text explaining the \`=\`-not-\`:\` separator rule (\`docs/gotchas.md\` documents Scaleway's API rejects colon-separated tags). \`create_user\`'s tags field has neither.

**Failure scenario:** an agent calling \`scaleway_iam_create_user\` passes an unbounded array of arbitrary-format tag strings (including ones using \`:\` as a separator) - this fails only via a late, less-clear server-side 400 instead of the fast, clear client-side validation every other tags field in this server provides.

**Suggested fix:** copy the \`.max(10)\` + locked-vocabulary description from \`update_user\`'s tags field (same file) into \`create_user\`'s.

gkrost:
Closes #50.

**Stack:** 8/9. Base is `fix/49-copy-object-bucket-name`.

## Summary
`create_user` tags had no `.max(10)` and no locked-vocabulary / `=` vs `:` description. `update_user` already did.

- Copy the cap and describe text onto create (complete tag set on the new user).

## Test plan


src/tools/users.ts
```diff
@@ -90,7 +90,17 @@ export function registerUsers(server: McpServer, config: Config) {
       inputSchema: {
         email: z.string().email().describe("The person's email address - receives the invitation."),
         type: z.enum(["guest", "member"]).default("guest").describe("User type. 'guest' default (inferred, unverified)."),
-        tags: z.array(z.string()).optional(),
+        tags: z
+          .array(z.string())
+          .max(10)
```
gkrost, line 95: Cap matches `update_user`'s `.max(10)` exactly (verified: 10 in both, no off-by-one), and the replace-not-merge wording is consistent between the two tools. Good.

One gap: nothing in this diff proves the cap actually rejects an over-limit array — the only `create_user` coverage in `scripts/smoke-test.mjs` is the invalid-email negative test a few lines below (which fires purely on `email`, never touches `tags`). Pushed a regression test on this branch (commit `c389804`) that calls `create_user` with 11 tags + a reserved `example.invalid` address and asserts `isError`, mirroring the existing invalid-email test's "schema validation fires before any email is sent" pattern.

gkrost (commented):
Reviewed against #58 (base) and the update_user tags field it mirrors.

- Cap matches update_user's `.max(10)` exactly - no off-by-one, no inconsistency between the two paths.
- Description text (locked vocabulary, `=` not `:`, non-merge caveat) is consistent between create_user and update_user.
- Gap found: no test proved the new cap actually rejects an over-limit array (see inline comment). Pushed a fix on this branch, commit c389804 - adds a create_user call with 11 tags + a reserved `example.invalid` address, asserting the MCP schema validation rejects it before any real invite could be sent. `npm run build` and `node --check scripts/smoke-test.mjs` both pass on the updated branch.

No other issues found in this diff.

PR #67, opened 2026-08-19
fix(s3): confirm-gate delete_bucket_encryption

### LINKED ISSUE #64: delete_bucket_encryption silently disables a live security control with no confirm gate, unlike its siblings
## What's wrong

`scaleway_s3_delete_bucket_encryption` has no `confirm` gate, unlike its siblings in the same file (`delete_bucket_tagging`, `delete_bucket_lifecycle`, `delete_bucket_policy` all require `confirm=true`) - despite silently disabling a live security control.

`scaleway_s3_get_bucket_encryption`'s own description (added when this feature was implemented) is explicit that this is not inert metadata: *"this is a REAL, toggleable setting... the console's Settings tab shows 'Encryption type' flip from 'Disabled' to 'SSE-ONE encryption...' immediately after this call."* `put_bucket_encryption`'s and `delete_bucket_encryption`'s descriptions confirm the same both ways. So removing it is a genuine security-posture downgrade (future uploads that don't explicitly request SSE stop getting it by default), the same "loosening direction" class of change that `scaleway_iam_update_security_settings` explicitly calls out as needing `confirm=true` ("loosening... weakens the org's security posture for everyone").

## Where

`src/tools/bucketConfig.ts`, `scaleway_s3_delete_bucket_encryption` (~line 458):

```ts
inputSchema: { bucket: bucketField, region: regionField },
annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: false, openWorldHint: true },
```

No `confirm` field, no check, and `destructiveHint: false` despite the sibling `delete_bucket_tagging` (a strictly lower-severity removal - labels, not a protection control) carrying `destructiveHint: true` and a `confirm` requirement.

## Why it matters

An agent asked to "clean up bucket config" or acting on a plausible-sounding instruction can silently turn off default encryption on a bucket with no confirmation step at all, where every comparably-scoped sibling in the same file requires one.

## Fix

Add a `confirm: z.literal(true)` field to `delete_bucket_encryption`'s schema, check it in the handler (same pattern as `delete_bucket_tagging`/`delete_bucket_lifecycle`), update the description to state the requirement, and set `destructiveHint: true`.

gkrost:
## Summary

- `scaleway_s3_delete_bucket_encryption` had no `confirm` gate, unlike its siblings `delete_bucket_tagging`/`delete_bucket_lifecycle`/`delete_bucket_policy` in the same file - despite silently disabling a live security control.
- The tool's own `get`/`put_bucket_encryption` descriptions are explicit this is a real, toggleable setting (live-verified via the console's "Encryption type" flip between Disabled/SSE), not inert metadata - so removing it is a genuine security-posture downgrade, the same "loosening" class of change `scaleway_iam_update_security_settings` already requires `confirm=true` for.
- Adds `confirm: confirmField` to the schema (the existing non-optional `z.literal(true)` pattern used by every sibling delete tool in this file - zod rejects a missing `confirm` at schema validation, no handler-side check needed) and sets `destructiveHint: true`.

## Test plan

- [x] `node --check scripts/smoke-test.mjs` passes
- [x] Added a negative smoke-test case (`delete_bucket_encryption` without `confirm` rejected) and updated the existing positive-path call to pass `confirm: true`, following the exact same pattern already used for `delete_bucket_tagging`/`delete_bucket_lifecycle` a few lines up in the same file
- [ ] Live smoke test (`node scripts/smoke-test.mjs`) - **not run**, no live Scaleway credentials available in this environment

Closes #64


src/tools/bucketConfig.ts
```diff
@@ -462,9 +462,11 @@ export function registerBucketConfig(server: McpServer, config: Config) {
     {
       title: "Delete Scaleway bucket encryption config",
       description:
-        "Remove the bucket's default-encryption configuration. Live-verified 2026-08-18: this reverts the console's 'Encryption type' back to 'Disabled' - a real change, not a no-op on inert metadata.",
-      inputSchema: { bucket: bucketField, region: regionField },
-      annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: false, openWorldHint: true },
+        "Remove the bucket's default-encryption configuration. Requires confirm=true: live-verified 2026-08-18, " +
+        "this reverts the console's 'Encryption type' back to 'Disabled' - a real security-posture downgrade (future " +
+        "uploads that don't explicitly request SSE stop getting it by default), not a no-op on inert metadata.",
+      inputSchema: { bucket: bucketField, region: regionField, confirm: confirmField },
```
gkrost, line 468: Confirmed: `confirmField` here is the exact same shared constant used by every sibling delete tool in this file (delete_bucket_tagging, delete_bucket_lifecycle, delete_bucket_policy all use it identically). This closes issue #64 correctly - matches the established pattern, not a bespoke one-off.

scripts/smoke-test.mjs
```diff
@@ -477,7 +477,8 @@ try {
   const enc = expectJson(await client.callTool({ name: "scaleway_s3_get_bucket_encryption", arguments: { bucket: cfgBucket } }), "get_bucket_encryption");
   if (!enc.rules?.some((r) => r.algorithm === "AES256")) { console.error("FAILED: encryption echo mismatch"); process.exit(1); }
   console.log("encryption get: AES256 echoed");
-  console.log("encryption delete:", text(await client.callTool({ name: "scaleway_s3_delete_bucket_encryption", arguments: { bucket: cfgBucket } })).slice(0, 80));
+  await expectError("scaleway_s3_delete_bucket_encryption", { bucket: cfgBucket }, "delete_bucket_encryption without confirm rejected");
```
gkrost, line 480: Good - this negative-path check (no confirm -> rejected) plus the updated positive-path call on the next line give this the same two-sided coverage every other confirm-gated tool's smoke test already has. Confirmed no other call site in the repo invokes this tool without confirm, so nothing else breaks.

gkrost (commented):
Reviewed against the codebase's established confirm-gate conventions before approving - not just skimming the diff.

Verified:
- `confirmField` (line 468) is the same shared `z.literal(true)` constant every sibling delete tool in this file already uses (`delete_bucket_tagging`, `delete_bucket_lifecycle`, `delete_bucket_policy`) - confirmed by grepping all 9 uses in `bucketConfig.ts`, this isn't a one-off pattern.
- The handler signature intentionally stays `async ({ bucket, region }) =>` without destructuring `confirm` - matches the rest of the codebase, since the MCP SDK rejects a call missing a `z.literal(true)` field at the schema-validation layer before the handler ever runs. Not a bug.
- Grepped the whole tree for other callers of `scaleway_s3_delete_bucket_encryption` - only `scripts/smoke-test.mjs` calls it, and this PR updates that call site too (both the negative no-confirm case and the positive confirm:true case). No other caller left broken.

Correct and complete. Approving.

PR #68, opened 2026-08-19
test: add put/get/delete_bucket_policy round-trip to smoke-test.mjs

### LINKED ISSUE #65: Bucket Policy tools (put/get/delete_bucket_policy) have zero coverage in smoke-test.mjs
## What's wrong

`scaleway_s3_put_bucket_policy`, `scaleway_s3_get_bucket_policy`, and `scaleway_s3_delete_bucket_policy` have zero coverage anywhere in `scripts/smoke-test.mjs` - not even a basic put -> get -> delete -> verify-gone round trip against the live API. Every other bucket-config feature in the same file/tool group (`bucketConfig.ts`) - tagging, CORS, versioning, website, visibility, lifecycle, encryption, object lock - gets exactly this kind of round-trip check inside the `try { ... } finally { teardown }` block starting around line 335. Bucket Policies (`bucketPolicies.ts`) are the one sibling feature missing from that block entirely.

This isn't a deliberate omission the way the SAML/SCIM live-testing gaps are (those are explicitly commented as intentionally avoided, to dodge a known empty-body incident - see `docs/gotchas.md`). Nothing in `bucketPolicies.ts` or the smoke test documents a reason bucket policies are skipped; it reads as an oversight.

## Why it matters

Bucket Policies are one of the most security-relevant primitives in this server - the tool's own description says a policy document "can grant `Principal *`" (anonymous/public access to a bucket) or "drop the caller's own access." That's a materially higher blast radius than, say, CORS rules or website config, both of which *are* live-tested. The actual PUT/GET/DELETE round trip against Scaleway's API - confirming the policy document that comes back from `get_bucket_policy` actually matches what was PUT, and that `delete_bucket_policy` really removes it (`get_bucket_policy` errors afterward) - has never been exercised, so a shape/serialization bug in this specific path (e.g. the bare-bucket-name-not-ARN quirk the tool's own description already warns about) would not be caught by this suite.

## Fix

Add a bucket-policy round-trip to the same `try`/`finally` config-bucket block in `scripts/smoke-test.mjs`: PUT a real (non-destructive, scoped-down) policy document, GET it back and verify the statement round-tripped, DELETE it, then GET again and verify it errors (no policy). Test-only change, no production code touched.

gkrost:
## Summary

- `scaleway_s3_put_bucket_policy`, `scaleway_s3_get_bucket_policy`, and `scaleway_s3_delete_bucket_policy` had zero coverage anywhere in `scripts/smoke-test.mjs` - not even a basic round trip. Every other bucket-config feature in `bucketConfig.ts` (tagging, CORS, versioning, website, visibility, lifecycle, encryption, object lock) gets a put -> get -> delete round trip inside the `try { ... } finally { teardown }` block for the throwaway config bucket; Bucket Policies were the one sibling missing.
- This isn't a deliberate omission the way the SAML/SCIM live-testing gaps are (those are explicitly commented as intentionally avoided in the source). Nothing documents a reason bucket policies are skipped - it reads as an oversight.
- Bucket Policies are one of the higher-blast-radius primitives here - `put_bucket_policy`'s own description says a document can grant `Principal *` or drop the caller's own access - so the real PUT/GET/DELETE round trip (confirming `get_bucket_policy` actually echoes what was PUT, and `delete_bucket_policy` really removes it) had never been exercised.

## What the new test does

Adds a bucket-policy block to the config-bucket lifecycle test: PUTs a real, minimal, non-destructive statement (a read-only grant to the smoke test's own already-fully-privileged Application, on a throwaway bucket deleted moments later - so it changes nothing materially), GETs it back and verifies the statement round-tripped (`Sid` + `Principal`), DELETEs it, then verifies `get_bucket_policy` errors afterward. Also exercises the bare-bucket-name-not-ARN `Resource` shape the tool's own description warns about.

Test-only change - no production code touched.

## Test plan

- [x] `node --check scripts/smoke-test.mjs` passes
- [ ] Live smoke test (`node scripts/smoke-test.mjs`) - **not run**, no live Scaleway credentials available in this environment. Verified via code review + syntax/build checks instead, per repo convention for this kind of change.

Closes #65


scripts/smoke-test.mjs
```diff
@@ -451,6 +451,40 @@ try {
...
+  };
+  expectJson(
+    await client.callTool({
+      name: "scaleway_s3_put_bucket_policy",
+      arguments: { bucket: cfgBucket, policy_json: JSON.stringify(bucketPolicyDoc), confirm: true },
+    }),
+    "put_bucket_policy",
+  );
+  const policyBack = expectJson(await client.callTool({ name: "scaleway_s3_get_bucket_policy", arguments: { bucket: cfgBucket } }), "get_bucket_policy");
+  if (policyBack.policy?.Statement?.[0]?.Sid !== "SmokeTestReadOnly" || policyBack.policy?.Statement?.[0]?.Principal?.SCW !== `application_id:${appId}`) {
```
gkrost, line 509: Confirmed this matches get_bucket_policy's real response shape (`{bucket, policy}`, policy = parsed JSON or null) - checked src/tools/bucketPolicies.ts directly rather than assuming. Good assertion.

scripts/smoke-test.mjs
```diff
@@ -451,6 +451,40 @@ try {
...
+    console.error(`FAILED: get_bucket_policy did not echo the applied statement: ${JSON.stringify(policyBack)}`);
+    process.exit(1);
+  }
+  console.log("bucket policy get: statement echoed (Sid + Principal match)");
+  console.log("bucket policy delete:", text(await client.callTool({ name: "scaleway_s3_delete_bucket_policy", arguments: { bucket: cfgBucket, confirm: true } })).slice(0, 80));
+  await expectError("scaleway_s3_get_bucket_policy", { bucket: cfgBucket }, "no bucket policy after delete");
```
gkrost, line 515: Good - this is what actually proves the delete worked (policy gone -> subsequent get errors), not just that the DELETE call itself returned success.

gkrost (commented):
Reviewed for correctness against the actual tool implementation, not just read for intent.

Verified:
- `appId` (used at line 466/480) is resolved at the top of the script (line 65, `const appId = expectJson(apps, "list_applications").applications[0].id`), well before this new block - in scope, not a forward reference.
- The assertion shape at line 480 (`policyBack.policy?.Statement?.[0]?.Sid` / `.Principal?.SCW`) matches `get_bucket_policy`'s actual handler exactly - it returns `{bucket, policy}` where `policy` is `res.Policy ? JSON.parse(res.Policy) : null` (src/tools/bucketPolicies.ts), so `.policy.Statement[0]` is the right path, not a guess.
- `Version: "2023-04-17"` + the `application_id:` Principal shape matches the documented gotcha in docs/gotchas.md exactly (AWS's 2012-10-17 version fails for this Principal type).
- The post-delete check (line 486, `expectError` on a subsequent `get_bucket_policy`) correctly proves the delete actually took effect, not just that the DELETE call returned 200.
- Grepped for interaction with the object-lock test that follows this block in the same try/finally - no shared state, no ordering conflict.

Correct and complete - this was a real, undocumented gap (bucket policies are one of the higher-blast-radius primitives here) and the round trip actually exercises the real API, not just a schema check.

PR #83, opened 2026-09-21
feat: fix all 7 open issues (#73 #75 #76 #77 #78 #80 #81) - S3 ops hardening batch

### LINKED ISSUE #73: AccessDenied on objects gives no hint that a Bucket Policy excludes this server's own principal: add a policy-aware error and an explain-access tool
## What happened (2026-09-21, v0.2.0)

Every object-level call against the project's real buckets failed with the same bare message:

```
scaleway_s3_list_objects         dev.zvg-backups -> Scaleway Object Storage error (AccessDenied): Access Denied
scaleway_s3_get_bucket_lifecycle dev.zvg-backups -> Scaleway Object Storage error (AccessDenied): Access Denied
scaleway_s3_list_objects         zvg-backups     -> (AccessDenied)
scaleway_s3_list_objects         zvg-files       -> (AccessDenied)
scaleway_s3_list_objects         dev.zvg-files   -> (AccessDenied)
```

Meanwhile `scaleway_s3_list_buckets` and **`scaleway_s3_get_bucket_policy` on the same buckets succeeded**, and the policies explain everything. Each one grants object access only to the owner `user_id` and one or two named `application_id`s: `dev.zvg-backups` → a `BackupWriter` application; `zvg-files` → `ProdAttachmentStore` + `LocalDevReadOnly`. **This server's own application is in none of them**, so the denial is correct: per-stage isolation doing its job. But getting there took four failed calls and a manual policy read, and the error text (`src/s3Client.ts:37`) names neither the principal nor the mechanism.

## Recommendation

1. **Enrich S3 `AccessDenied`.** On an object-level denial, read the bucket's policy (the server evidently may). Say which principal the call ran as (this server's `application_id`), and whether the bucket has a policy that does not list it. For example: *"AccessDenied: bucket dev.zvg-backups has a Bucket Policy granting object access only to user_id:edd2…, application_id:86f2… — this server runs as application_id:&lt;own&gt; and is not listed."*
2. **A read-only `scaleway_whoami` / `scaleway_s3_explain_access(bucket, action)` tool.** It resolves this server's own Application (`src/tools/applications.ts` already does that for the self-delete guards, #63) and evaluates a bucket's policy statements against it: allowed or denied, and by which `Sid`.

## Considered, not recommended
Multiple credential profiles, so the server could "just use" each stage's key. That would put every stage's credential into one process, which is exactly the isolation the policies exist to enforce. Explaining the denial is enough.

### LINKED ISSUE #75: copy_object fails above 5 GB (single CopyObject); no bucket-to-bucket copy with ignore-existing and dry-run
Two gaps that showed up while planning a prod → dev attachment sync (gkrost/immo-service#528):

## 1. `scaleway_s3_copy_object` fails above 5 GB
It issues one `CopyObjectCommand` (`src/tools/objects.ts:353`). S3's single-request copy has the same 5 GB ceiling as a single PUT, so it cannot copy, for example, the 6.8 GB ODS backup dump. Above the threshold, use a multipart copy (`CreateMultipartUpload` → `UploadPartCopy` ranges → `Complete`, abort on failure). The tool already performs a `HEAD` for its overwrite check, so the size is free.

## 2. No bucket-to-bucket copy
Syncing the objects missing from one bucket into another (`rclone copy --ignore-existing` semantics, **server-side**) takes thousands of `copy_object` calls today. Proposal: `scaleway_s3_copy_prefix(source_bucket, dest_bucket, prefix?, ignore_existing=true, dry_run=true, max_objects)`.
- **Dry run by default**: returns how many objects and bytes *would* be copied, plus a sample of keys.
- Paginates both listings, copies server-side, and skips existing keys (optionally comparing size/etag).
- Bounded per call with a resume token; reports `copied / skipped / failed`.

A note on access (#73): a server-side copy needs one principal with read on the source **and** write on the destination. When stages are isolated by Bucket Policy, that principal only exists after an explicit and ideally temporary grant (#77).

### LINKED ISSUE #76: No read-only bucket stats (count, bytes, oldest/newest per key stem): retention drift goes unnoticed
Retention and size questions currently mean paging raw `list_objects` output by hand. A read-only summary would answer them in one call.

## Proposal: `scaleway_s3_bucket_stats(bucket, prefix?, group_by='key-stem'|'prefix'|none)`
Returns object count, total bytes and oldest/newest `LastModified`, overall and per group. A group can be a key stem with its trailing timestamp stripped, e.g. `zvg_pg_2026-09-21T01-30-00.dump` → `zvg_pg`. It paginates internally, caps the object count, and reports a `truncated` flag.

## The finding that motivated it (2026-09-21)
I listed `dev.zvg-backups` from the host with the app's own credential, since this server has no object access there:

| stem | objects | oldest | newest | bytes |
|---|---:|---|---|---:|
| zvg_pg | 39 | 2026-08-13 | 2026-09-21 | 30.8 GB |
| zvg_rawdetail | 39 | 2026-08-13 | 2026-09-21 | 5.6 GB |
| zvg_extractions | 38 | 2026-08-14 | 2026-09-21 | ~0 |
| opendata_pg | 2 | 2026-09-04 | 2026-09-05 | 9.7 GB |

The app's code comment says the bucket must carry a **30-day** expiration rule, yet objects are **39 days** old. This tool would have exposed that drift. A companion **`scaleway_s3_audit_lifecycle`** would flag buckets with no expiration or abort-incomplete-multipart rule, or whose oldest object is older than the rule allows, so nobody has to know to go looking.

### LINKED ISSUE #77: Bucket Policy edits are full-replace only; temporary cross-bucket grants have no statement-level add/remove and no expiry
Granting temporary cross-stage access is a recurring operator task, and with this server it is manual and error-prone today.

## Current state
- `scaleway_s3_put_bucket_policy` is **full-replace**; its own description says to GET, edit the `Statement` array, then PUT the merged result. One omitted statement silently revokes an existing grant, such as the app's own `BackupWriter`.
- API keys already support `expires_at` (`src/tools/apiKeys.ts:23`), but **a Bucket Policy statement has no expiry**, so a "temporary" grant outlives its key unless someone remembers to remove it.

## Proposal
1. **Statement-level tools.**
   - `scaleway_s3_add_bucket_policy_statement(bucket, statement)`, which refuses a duplicate `Sid`.
   - `scaleway_s3_remove_bucket_policy_statement(bucket, sid)`.

   Both do the read-modify-write server-side, with an ETag/precondition check or at least a re-read-and-compare. Removing the last statement should be refused or require `confirm`.
2. **A guided temporary grant**: `scaleway_s3_grant_temporary_access(application_id, bucket, actions, expires_at, sid_prefix='tmp-')` adds a `tmp-…` statement. A matching sweep removes `tmp-*` statements past their `expires_at` (recorded in the Sid or a tag). It could be a tool of its own, `scaleway_s3_revoke_expired_grants`, or a check run on every policy tool call.

## Motivating case (gkrost/immo-service#528)
A server-side copy of PDFs from `zvg-files` to `dev.zvg-files` needs one principal with read on the prod bucket and write on the dev bucket. Today that means: create an app, mint a key with `expires_at`, then GET, merge and PUT two bucket policies. Then run the copy, and remember to reverse all of it by hand.

### LINKED ISSUE #78: No tools to list or abort incomplete multipart uploads (billed, invisible parts)
There's no way to see or clean up **incomplete multipart uploads**: parts you pay storage for that no object listing shows.

## Proposal
- `scaleway_s3_list_multipart_uploads(bucket, prefix?)` (read-only): upload id, key, when it was started, and the part count and bytes via `ListParts` where that's cheap.
- `scaleway_s3_abort_multipart_upload(bucket, key, upload_id, confirm)`: guarded like the other deletes.

## Why now
gkrost/immo-service#529 introduces multipart uploads for backup archives over 1 GiB. The writer's bucket policy has no `s3:AbortMultipartUpload`, so a failed upload leaves its parts behind. #74 covers preventing that; without these tools nobody can tell whether it has already happened, or clean it up once it has.

### LINKED ISSUE #80: Lifecycle rules are full-replace only: adding one rule (e.g. abort-MPU) silently drops the others; add rule-level add/remove and a diff/dry-run
`scaleway_s3_put_bucket_lifecycle` is **FULL-REPLACE**: the list you pass becomes the bucket's entire rule set (`src/tools/bucketConfig.ts:357`). There is no way to add or remove **one** rule. That's the same trap #77 describes for Bucket Policies, and it bites #74 directly.

## Why it matters
Suppose the backup buckets carry a 30-day expiration rule and you want to add `AbortIncompleteMultipartUpload` (gkrost/immo-service#526/#529). A caller who PUTs only the new abort rule **silently deletes the expiration rule**, and the backups then never expire. Nothing in the call warns about it. The safe path today is get → merge by hand → put, and every caller has to remember that, whether human or agent.

## Proposal
- `scaleway_s3_add_lifecycle_rule(bucket, rule)`: server-side get → append → put. It refuses a duplicate `ID`, and keeps the existing destructive-rule `confirm` gate for expiration rules.
- `scaleway_s3_remove_lifecycle_rule(bucket, id, confirm)`: removes one rule by `ID`. Removing an expiration rule changes retention, so it needs `confirm`.
- **Show a diff** in `put_bucket_lifecycle` (and `put_bucket_policy`, #77): which rules are added, removed or changed compared with the current configuration. Offer an optional `dry_run` that returns the diff without writing.

Depends on #74 for the abort element, but is independent of it.

### LINKED ISSUE #81: Read-only security audit: buckets without policy, cross-stage/dangling principals, keys without expiry or orphaned after rotation
The server can already **read** everything a basic Scaleway access audit needs: `list_buckets`, `get_bucket_policy` (which works even where object access is denied), `list_applications`, `list_api_keys` and `list_policy_rules`. Today a person has to cross-reference those outputs by hand. A single read-only tool could do it, and the first manual pass on 2026-09-21 already found real issues.

## Proposal: `scaleway_security_audit(scope?)` (read-only)
Returns findings grouped by severity. Each finding names the resource and the reason.

**Buckets**
- A bucket with **no Bucket Policy**, where access is decided by IAM alone. Any application holding a project-scope Object Storage permission set can then reach it, including delete if it holds `ObjectStorageObjectsDelete`.
- Principals in a policy that **no longer exist** (deleted application or user ids).
- **Cross-stage principals**: a principal whose application name or description says "dev"/"local" granted on a bucket without a `dev.` prefix, and vice versa. This is a heuristic, so the finding is labelled as one.
- Statements granting `*` or delete/abort actions to a non-owner principal.

**API keys**
- Keys with **no expiry** (`expires_at: null`).
- Applications with **more than one active key**, and descriptions saying a key was replaced, rotated or "lost" while the old key still exists.
- This server's **own** key, if it has no expiry. Worth a startup WARN as well.

## What the manual pass found (2026-09-21, from this server's own list/get output)
- `mcp-test-drive-20260818` has **no bucket policy** (`NoSuchBucketPolicy`).
- `zvg-files` (prod attachments) grants `s3:GetObject`/`s3:ListBucket` to application `8ea2a3f5…`, whose keys are described as *"local win11 dev stack - read-only zvg-files"*. A local dev machine has read access to production data. It may be intended, but it's exactly the kind of grant an audit should surface.
- **3 of 11 keys never expire**: `SCWP1KZF…` and `SCW4DZFJ6…` (both `8ea2a3f5`, the local dev stack), and `SCW5H09W…` (this server's own operational key).
- `SCWP1KZF…` is still active although its sibling key's description says the *"prior 2026-08-12 key's secret lost with the torn-down local stack"*. That's an orphaned, never-expiring credential with read access to prod attachments. **It should be revoked**; that's an owner action, not something this tool should do.

## Out of scope
Changing anything. The tool reports, and every fix stays an explicit, confirm-gated call to the existing tools.

gkrost:
## Summary

Implements all seven open enhancement issues in one reviewable series (one commit per issue). All changes follow the repo's established patterns: confirm-gated writes, read-first guard checks with the accepted read-then-write race window, instructive client-side validation for the documented Scaleway traps, and best-effort explainers that never mask the original error.

**Closes #73, closes #75, closes #76, closes #77, closes #78, closes #80, closes #81**

## What's in here

| Commit | Issue | What it adds |
|---|---|---|
| 1 | #73 | Every S3 tool's `AccessDenied` now reads the bucket's policy and says which principal ran and whether the policy lists it; new read-only `scaleway_s3_explain_access(bucket, action)` names the allowing/denying Sid |
| 2 | #78 | `scaleway_s3_list_multipart_uploads` (+ optional per-upload part count/bytes) and confirm-guarded `scaleway_s3_abort_multipart_upload` |
| 3 | #76 | `scaleway_s3_bucket_stats` (count/bytes/age per key-stem/prefix group) and `scaleway_s3_audit_lifecycle` (no-config / no-expiration / no-abort-MPU / retention-drift findings) |
| 4 | #77 | `add/remove_bucket_policy_statement` (merge, duplicate-Sid refusal), `grant_temporary_access` (epoch encoded in the Sid), `revoke_expired_grants` (preview / confirm sweep of lapsed `tmp-*`) |
| 5 | #80 | `add_lifecycle_rule` / `remove_lifecycle_rule` (server-side merge; SDK rules round-trip untouched) + added/removed/changed/unchanged diff and `dry_run` on both `put_bucket_lifecycle` and `put_bucket_policy` |
| 6 | #75 | `copy_object` switches to multipart copy above 5 GiB (512 MiB ranged `UploadPartCopy`, aborted on failure) + `scaleway_s3_copy_prefix` (`rclone copy --ignore-existing` semantics, dry-run default, bounded with a continuation token) |
| 7 | #81 | Read-only `scaleway_security_audit` (no-policy buckets, dangling principals, cross-stage heuristics, `Principal "*"`, never-expiring / possibly-orphaned keys) + stderr startup WARN when this server's own key never expires |

## Verification

- `tsc` build green at every commit (CI runs the same).
- Offline boot test: server starts with dummy credentials; 130 tools registered, all 13 new names present.
- `scripts/smoke-test.mjs` extended with a zero-residue section covering the new tools: copy_prefix dry-run/run/skip-existing/overwrite-guard, bucket_stats key-stem grouping, audit_lifecycle's rule-less flag, the full lifecycle merge story (abort-only add without confirm, dry-run diff, duplicate-ID refusal, expiration-removal confirm gate), statement tools (add, duplicate-Sid refusal, epoch-Sid temporary grant, sweep preview→confirm on an already-lapsed grant incl. "last statement deletes the policy", explain_access verdict, last-statement refusal), and a read-only `security_audit` call. **Not run against the live account in this PR** - it mutates live IAM/buckets by design, so it should be run by hand (`node scripts/smoke-test.mjs`) before merge, as usual.
- The >5 GiB multipart-copy path and the `User`-principal dangling check are the two paths the smoke test cannot cheaply exercise live (needs a >5 GB object / policies granting user_ids); both are implemented against the documented S3 semantics and flagged here for the reviewer.

## Notes for review

- Read-modify-write tools (policy statements, lifecycle add/remove, revoke sweep) carry the same accepted read→write race as the existing abort-only lifecycle gate - Scaleway exposes no ETag to precondition on; tool results say so.
- The audit's cross-stage and orphaned-key checks are labelled HEURISTIC in their findings, per the issue.
- `remove_lifecycle_rule` refuses to remove the bucket's last rule and points at `delete_bucket_lifecycle`; `revoke_expired_grants` deletes the whole policy document when a sweep empties it (an empty `Statement` array is not a valid policy).


gkrost (commented):
Verified review findings and fixed them in 953b86a7d2194a28169ae91869dde1d89bff9781.

Confirmed defects addressed:
- bucket-level and object-level policy resources were treated as interchangeable; multipart object actions such as UploadPart/ListParts were also misclassified;
- user-owned API keys were represented as access_key rather than user_id, and an IAM lookup failure could become a false “policy excludes this server” or false dangling-principal conclusion;
- copy_prefix could skip the remainder of a source page at max_objects and ignored that cap in dry-run; cross-bucket AccessDenied context named only the destination;
- ListParts undercounted uploads with more than 1,000 parts; the copy threshold used 5 GiB instead of the documented 5 GB limit;
- lifecycle audit ignored date expiration, made prefix/filter-scoped retention look like global drift, and aborted all-bucket audits on one unreadable lifecycle config;
- the security audit missed wildcard destructive grants and accepted arbitrary regions.

Validation: npm ci; npm run build; focused policy evaluator boundary assertions; git diff --check. The live smoke script was intentionally not run: it creates and deletes live IAM and Object Storage resources.

